PR #2, opened 2023-09-26
add better styled typings

jantimon:
this pr improves typings for `styled` so that `styled.button` will allow all valid `<button>` props like `onClick`:

shot-jT9euzpi


Mad-Kat (approved):
LGTM

PR #5, opened 2023-10-01
move babel plugin to separate file

jantimon:
moving the babel plugin into a stand alone plugin allows to use it for jest.

right now it would NOT work as jest plugin because it is 1:1 version of the previous implementation.

to be useful in a jest / vite environment the plugin would require a flag which would:

- not inejct the import
- would not use the injected import but rather a string like "yak-styles"

we should also add a test in our example package

### DIFF
--- a/packages/next-yak/loaders/__tests__/tsloader.test.ts
+++ b/packages/next-yak/loaders/__tests__/tsloader.test.ts
@@ -20,6 +20,12 @@ const loaderContext = {
   getOptions: () => ({
     configPath: "/some/special/path/config",
   }),
+  async: () => (err, result) => {
+    if (err) {
+      throw err;
+    }
+    return result;
+  }
 };
 
 describe("tsloader", () => {
--- a/packages/next-yak/loaders/tsloader.cjs
+++ b/packages/next-yak/loaders/tsloader.cjs
@@ -3,13 +3,13 @@ const babel = require("@babel/core");
 const { resolve } = require("path");
 
 /**
- * Loader for typescript files that use yacijs, it replaces the css template literal with a call to the 'styled' function
+ * Loader for typescript files that use yak, it replaces the css template literal with a call to the 'styled' function
  * @param {string} source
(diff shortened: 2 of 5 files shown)


packages/next-yak/loaders/tsloader.cjs
```diff
@@ -1,240 +1,56 @@
...
-  const { types: t } = babel;
-  const fileName = path.basename(this.resourcePath).replace(/\.tsx?/, "");
-  // parse source with babel
+  // Parse source with babel and pass options to the Babel plugin
   const result = babel.transformSync(source, {
-    filename: this.resourcePath,
+    filename: resourcePath,
     configFile: false,
-    // Only for parsing - will be removed once moved to a swc or babel plugin
+    // Use the Babel plugin you created here and pass the required options
```
Mad-Kat, line 32: Can be removed

packages/next-yak/loaders/__tests__/tsloader.test.ts
```diff
@@ -20,6 +20,12 @@ const loaderContext = {
   getOptions: () => ({
     configPath: "/some/special/path/config",
   }),
+  async: () => (err, result) => {
```
Mad-Kat, line 23: Why do we need this here?

Mad-Kat (changes requested):

packages/next-yak/loaders/tsloader.cjs
jantimon: danke :)

packages/next-yak/loaders/__tests__/tsloader.test.ts
jantimon: webpack loaders can't return source maps with a `return` statement.

it works only if we use the `async` method: https://webpack.js.org/api/loaders/#thisasync

Mad-Kat (approved):
Thanks for the changes

PR #6, opened 2023-10-01
recursive prop evalutation

jantimon:
backport from our intern repository


Mad-Kat (approved):
A test would be nice

jantimon:
good point - added a test 👍

Mad-Kat (approved):
Thank you for the additional test

PR #7, opened 2023-10-03
Add .attrs to styled

Mad-Kat:
# Description
This PR adds the ability to use `.attrs` on `styled`.

It follows the API of styled-components very closely with 2 exceptions:
- Multiple `.attrs` calls on the same component don't work. But `.attrs` works on a wrapped component if the wrapped component has already used `.attrs`
```jsx
const Button = styled.button.attrs().attrs()`` // doesn't work

const Link = styled.a.attrs()``
const WrappedLink = styled(Link).attrs()`` // works
const OtherWrappedLink = styled(Link).attrs().attrs()`` // doesn't work
```
- `as` property doesn't change the underlying html element
```jsx
const Link = styled.div.attrs({as: "a"})``

// results in 
// <div></div>

// and not in
// <a></a>
```

This is open for debate on whatever we should add this functionality or not. I'm leaning towards not implementing it, but I don't have a strong opinion on that.

## Remarks

Overriding props works, but I added an "opaque" property to keep track if it's the first invocation of `.attrs` or if it's a wrapped invocation to override the "original" `.attrs`. I'm not really sure if this is the right way to go. It adds more runtime complexity & time, but we follow the API of `styled-components` more closely.

```jsx
const Parent = styled.button.attrs(() => ({
    type: "button",
    tabIndex: 0,
}))``;

const Child = styled(Parent).attrs(() => ({
    type: "submit",
}))``;

// results in
<button
    tabIndex={0}
    // type is "submit" as it gets overwritten
    type="submit"
/>
```


jantimon:
can we exclude the tests from the typescript compiling?

packages/next-yak/runtime/styled.tsx
```diff
@@ -12,39 +12,106 @@ import React from "react";
...
+              (props.className ? props.className + " " : "") +
+              runtimeStyles.className
+            }
+          />
+        );
+      };
+    },
+    {
+      attrs: (attrsProps) => {
+        return (styles, ...values) => {
```
jantimon, line 75: would it be possible to call `this` here?

e.g. 

```tsx
return (props) => {
  const Component = this;
  const newProps = typeof attrsProps === "function"
                ? (attrsProps as Function)(props)
                : attrsProps;
  const combinedProps = combineProps(props, newProps);
  return <Component {combinedProps} >
}
```

Mad-Kat: I'm not sure, as the `props` from the component get passed in last and we need to transform them first.

jantimon (approved):

PR #29, opened 2023-11-13
codspeed use tsup

jantimon:


.github/workflows/codspeed.yml
```diff
@@ -35,9 +35,11 @@ jobs:
...
       - name: Install dependencies
         run: pnpm install
+      - name: Build benchmark suite
+        run: cd packages/benchmark && pnpm run codspeed 
       - name: Run benchmarks
         uses: CodSpeedHQ/action@v1
         with:
           working-directory: "packages/benchmark"
-          run: pnpm run codspeed
+          run: NODE_ENV=production node ./codspeed/dist/index.bench.mjs
```
adriencaccia, line 44: Hey @jantimon, we can try enabling source maps here

```suggestion
          run: NODE_ENV=production node --enable-source-maps ./codspeed/dist/index.bench.mjs
```

PR #44, opened 2024-01-15
Nested CSS Expressions

rvetere:


jantimon:
Todos:



- [ ] ts-loader - Allow using expressions inside nestings (remove `Expressions are not allowed inside nested selectors` error in those cases)
- [ ] ts-loader - Add `styled.` component class name even for nested css e.g.:

```tsx
css`
  ${x > 0.5 && css`
    color: blue;
  `}
`
```

jantimon (approved):

PR #63, opened 2024-02-07
inline constants

jantimon:


packages/next-yak/loaders/__tests__/cssloader.test.ts
```diff
@@ -503,16 +503,39 @@ const Component2 = styled.div\`
...
     expect(
       await cssloader.call(
         loaderContext,
         `
 import { styled, keyframes, css } from "next-yak";
 
-const color = "green";
-const duration = "1s";
-const easing = "ease-out";
+const color = new Token("green");
```
Mad-Kat, line 536: Do we need those values wrapped in a token?

packages/next-yak/loaders/__tests__/tsloader.test.ts
```diff
@@ -491,7 +515,7 @@ const headline = css\`
         `
         import { styled, css } from "next-yak";
 
-        const $red = "#E50914";
+        const $red = new Token("#E50914");
```
Mad-Kat, line 518: Do we need the Token?

Mad-Kat (commented):
I don't think we need the wrapping of the values in a `new Token`, but I'm maybe mistaken :)

What do you think?

packages/next-yak/loaders/__tests__/cssloader.test.ts
jantimon: yes otherwise they would be inlined - (so the test result would change)

packages/next-yak/loaders/__tests__/tsloader.test.ts
jantimon: yes otherwise they would now be inlined - (so the test result would change)

packages/next-yak/loaders/__tests__/cssloader.test.ts
Mad-Kat: Ah I understand :)

LGTM

Mad-Kat (approved):

PR #64, opened 2024-02-07
Extract css units in css expressions automatically

### LINKED ISSUE #54: allow css unit postfixes
next-yak converts expressions compiles runtime values like `padding: ${size}rem;` to `padding: var(--value)rem;`  
unfortunately this is invalid css. the unit must be part of the variable value e.g. `padding: var(--value);`

right now the user has to write `padding: ${size + "rem"};` which is not intuitive  
therefore we should support `padding: ${size}rem;` by trying to detect the unit automatically

Cases:

```ts
const case1 = css`${4}px`

const case2 = css`${({$indent}) => $indent > 4 ? 10 : $indent}px`

const case3 = css`${({$indent}) => {
  if ($indent > 4) return 10;
  return $indent
}}px`
```

rvetere:
- [x] Only solve complex arrow functions with the "unitPostFix" helper (BlockStatement inside body)
- [x] Cleanup code - simplify transform expression function, Identifier is not needed anymore we think
- [x] Use the Set pattern for helpers everywhere
- [x] Add more testcases (check if all have valid css :D)
- [x] Add better titles to testcases

fixes #54


packages/next-yak/loaders/babel-yak-plugin.cjs
```diff
@@ -366,6 +367,12 @@ module.exports = function (babel, options) {
               .varIndex++}`;
             // expression: `x`
             // { style: { --v0: x}}
+            const cssUnit =
+              quasis[i + 1] && extractCssUnit(quasis[i + 1].value.raw);
```
Mad-Kat, line 371: What happens if you have more than one unit in your quasi?
e.g.
```
margin: ${x}rem;
padding: 10px;
```

In this case your quasi would be `rem;\r\npadding: 10px;` and your detection wouldn't work, or would it?

packages/next-yak/loaders/__tests__/tsloader.test.ts
Mad-Kat, line 1: Thank you for the test. I would love to have some more tests, that cover more cases that can happen

Mad-Kat (changes requested):
Thank you for the PR :)
It goes in a good direction, but I think it's still missing some crucial cases that can happen.

packages/next-yak/loaders/babel-yak-plugin.cjs
```diff
@@ -366,6 +367,12 @@ module.exports = function (babel, options) {
...
+            if (cssUnit) {
+              // @ts-expect-error TODO: fix this
+              expression.value = expression.value + ` + "${cssUnit}"`;
```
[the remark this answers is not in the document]
Mad-Kat: What do you mean by that? Do you mean that we should know the compile time value at this point?

packages/next-yak/loaders/babel-yak-plugin.cjs
rvetere: oooh good catch, i didn't thought about that case yet ^^

rvetere: should be more bulletproof now by splitting on ";"

jantimon:
please add tests for those cases:

```ts
const case1 = css`${4}px`

const case2 = css`${({$indent}) => $indent > 4 ? 10 : $indent}px`

const case3 = css`${({$indent}) => {
  if ($indent > 4) return 10;
  return $indent
}}px`
```

One idea if we cant fix case3 (or even more complicated cases):

```ts
({$indent}) => {
  if ($indent > 4) return 10;
  return $indent
}
```

for properties an expression is a function which returns a number or string

we could add a helper which executes this function and appends the unit afterwards e.g.:

```tsc
const __yakAppendUnit = ( baseFunction, unit ) => {
   // run a new function which accepts the same args as `baseFunction`:
   return (...args) => {
      // execute the baseFunction
      const baseFunctionResult = baseFunction(...args);
      // add the unit
      return baseFunctionResult + unit
  }
}
```

that way we could solve even the most complicated cases:

```ts
const case3 = css`${__yakAppendUnit(

({$indent}) => {
  if ($indent > 4) return 10;
  return $indent
}

, "px")}`
```

jantimon (changes requested):

packages/next-yak/loaders/babel-yak-plugin.cjs
```diff
@@ -356,23 +377,38 @@ module.exports = function (babel, options) {
...
-            if (!cssVariablesInlineStyle) {
-              cssVariablesInlineStyle = t.objectExpression([]);
-            }
             if (!cssVariablesInlineStyle) {
               cssVariablesInlineStyle = t.objectExpression([]);
             }
             const cssVariableName = `--🦬${getHashedFilePath(state.file)}${this
               .varIndex++}`;
+
+            const isAnimationExpression =
```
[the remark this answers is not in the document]
rvetere: solved it in a new way, we need to further discuss this

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,80 @@
...
+ * Extracts the css unit from a css string and checks if it is a valid CSS unit
+ *
+ * @param {string} quasi
+ */
+function extractCssUnit(quasi) {
+  // Regex to match units composed of [a-z] or exactly "%"
+  const regex = /^([a-z]+|%)/;
+
+  // Clean the input string by removing whitespace, line breaks, and semicolons
+  const [firstCssString] = quasi.trim().split(";");
```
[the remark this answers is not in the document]
rvetere: let's discuss this shortly

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,80 @@
...
+            runtimeInternalHelpers,
+            expression: callExpression,
+          };
+        }
+      }
+
+      const cssUnitLiteral = t.stringLiteral(cssUnit);
+      const binaryExpression = t.binaryExpression(
+        "+",
+        expression,
```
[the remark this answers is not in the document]
rvetere: re-written

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,80 @@
...
+ */
+const handleCssUnitInExpression = (nextQuasi, expression, t) => {
+  const runtimeInternalHelpers = new Set();
+  if (
+    expression.type === "ArrowFunctionExpression" ||
+    expression.type === "NumericLiteral" ||
+    expression.type === "BinaryExpression" ||
+    expression.type === "Identifier"
+  ) {
+    const cssUnit = nextQuasi && extractCssUnit(nextQuasi.value.raw);
```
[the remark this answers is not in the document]
rvetere: done

packages/next-yak/loaders/__tests__/tsloader.test.ts
```diff
@@ -779,4 +779,124 @@ const Button = styled.button\`
...
+          }) => (index * 30) + \\"deg\\",
+          \\"--\\\\uD83E\\\\uDDAC18fi82j1\\": ({
+            index
+          }) => (-index * 30) + \\"deg\\"
+        }
+      });"
+    `);
+  });
+
+  it("should detect expressions with units in complex arrow functions and wrap them with __yak_unitPostFix helper", async () => {
```
jantimon, line 846: can't see any `__yak_unitPostFix`

packages/next-yak/loaders/babel-yak-plugin.cjs
```diff
@@ -128,10 +132,26 @@ module.exports = function (babel, options) {
...
       this.topLevelConstBindings = new Map();
+      this.runtimeInternalHelpers = new Set();
     },
     visitor: {
-      Program(path) {
-        this.topLevelConstBindings = getConstantValues(path, t);
+      Program: {
+        enter(path, state) {
+          this.topLevelConstBindings = getConstantValues(path, t);
+          // Initialize a state flag to track whether the import should be added
```
jantimon, line 141: ?

packages/next-yak/loaders/babel-yak-plugin.cjs
```diff
@@ -128,10 +132,26 @@ module.exports = function (babel, options) {
...
+        },
+        exit(path, state) {
+          // Check the flag and add the import at the end of processing
```
jantimon, line 144:
```suggestion
          // After processing the file add imports for all injected runtime helpers
```

packages/next-yak/loaders/babel-yak-plugin.cjs
```diff
@@ -356,23 +378,34 @@ module.exports = function (babel, options) {
...
-            }
             if (!cssVariablesInlineStyle) {
               cssVariablesInlineStyle = t.objectExpression([]);
             }
             const cssVariableName = `--🦬${getHashedFilePath(state.file)}${this
               .varIndex++}`;
+
+            const cssUnit =
+              quasis[i + 1] && extractCssUnit(quasis[i + 1].value.raw);
+            const transformedExpression = cssUnit
```
jantimon, line 389: very readable 👍👍
even better than my idea

one really tiny thing - could you please find a more concrete name than "handle"?

packages/next-yak/loaders/cssloader.cjs
```diff
@@ -234,11 +239,19 @@ module.exports = async function cssLoader(source) {
...
+      let removeCssUnit = false;
       for (let i = 0; i < quasis.length; i++) {
         const quasi = quasis[i];
         const expression = path.node.quasi.expressions[i];
         // loop over all quasis belonging to the same css block
         const type = quasiTypes[i];
+
+        let code = unEscapeCssCode(quasi.value.raw);
+        if (removeCssUnit) {
+          code = code.replace(/^([a-z]+|%)/, "");
```
jantimon, line 251: could we explain why we need to remove the css unit?

sth like:

```suggestion
          // a quasi might start with a css unit e.g. css`margin: ${value}px 0;`
          // the css unit will be part of the typescript runtime code and must
          // must be removed from the css code as `var(--foo)px` would be invalid css
          code = code.replace(/^([a-z]+|%)/, "");
```

packages/next-yak/loaders/lib/__tests__/extractCssUnit.test.ts
```diff
@@ -0,0 +1,39 @@
+import extractCSSUnit from "../extractCssUnit.cjs";
+import { describe, it, expect } from "vitest";
+
+describe("extractCssUnit", () => {
+  it("should handle all possible cases", () => {
+    const testCases = [
+      " ", // => null
+      ";\n  display: block;\n  ", // => null
+      " ease-out 1s infinite reverse both paused slidein;\n", // => ease-out
```
jantimon, line 9: this is wrong - ease-out is not a css unit but the next value

packages/next-yak/loaders/lib/__tests__/extractCssUnit.test.ts
```diff
@@ -0,0 +1,39 @@
...
+    ];
+
+    const result = testCases.map((cssString) => {
+      return extractCSSUnit(cssString);
+    });
+    expect(result).toMatchInlineSnapshot(`
+      [
+        null,
+        null,
+        "ease",
```
jantimon, line 27:
```suggestion
        null,
```

packages/next-yak/loaders/lib/__tests__/extractCssUnit.test.ts
```diff
@@ -0,0 +1,39 @@
...
+      "px;\n        z-index: ", // => px
+      ";\n        transform: translateX(", // => null
+      "px) translateY(", // => px
+      "px);\n        font-family: 'Arial', sans-serif;\n     ", // => px
+      "deg)\n          translate(0, -88px) rotate(", // => deg
+      "deg);\n     ", // => deg
+    ];
+
+    const result = testCases.map((cssString) => {
+      return extractCSSUnit(cssString);
```
jantimon, line 21: maybe this could improve the readability of the result as we always see input -> output

```suggestion
      return { code: cssString, unit: extractCSSUnit(cssString) };
```

packages/next-yak/loaders/lib/extractCssUnit.cjs
```diff
@@ -0,0 +1,26 @@
...
+ * ";\n        transform: translateX("                      => null
+ * "px) translateY("                                        => px
+ * "px);\n        font-family: "Arial", sans-serif;\n     " => px
+ * "deg)\n          translate(0, -88px) rotate("            => deg
+ * "deg);\n     "                                           => deg
+ *
+ * @param {string} quasi
+ */
+function extractCssUnit(quasi) {
+  const nextQuasi = quasi.includes(";") ? quasi.split(";")[0] : quasi;
```
jantimon, line 20: why do we need this line?

packages/next-yak/loaders/lib/extractCssUnit.cjs
```diff
@@ -0,0 +1,26 @@
+/**
+ * Extracts the css unit from a css string and checks if it is a valid CSS unit
```
jantimon, line 2:
```suggestion
 * Extracts the css unit from the start of a css quasi e.g.:
 *
 * ```js
 *   // in the AST this code would have two quasis: "margin: " and "px 0 14px;"
 *   css`margin: ${size}px 0 14rem;` 
 * ```
```

packages/next-yak/loaders/lib/extractCssUnit.cjs
```diff
@@ -0,0 +1,26 @@
...
+ *
+ * example:
```
jantimon, line 4: maybe skip these examples

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,85 @@
+const CSS_ANIMATION_UNITS = [
```
jantimon, line 1: please remove

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,85 @@
...
+  "running",
+  "alternate",
+  "slidein",
+  "slideout",
+  "jump",
+  "start",
+  "end",
+  "normal",
+];
+const FORBIDDEEN_UNITS = [...CSS_ANIMATION_UNITS];
```
jantimon, line 21: please remove

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,85 @@
...
+  runtimeInternalHelpers,
+  t
+) => {
+  if (
+    expression.type === "ArrowFunctionExpression" ||
+    expression.type === "NumericLiteral" ||
+    expression.type === "BinaryExpression" ||
+    expression.type === "Identifier"
+  ) {
+    if (!FORBIDDEEN_UNITS.includes(cssUnit)) {
```
jantimon, line 43: why forbidden units?

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,85 @@
...
+      if (expression.type === "ArrowFunctionExpression") {
+        /**
+         * Functions that are not directly returing a value are wrapped with a helper function:
```
jantimon, line 46: I guess this comment belongs to the else case - could you please move it and also add a comment here for the __yak_unitPostFix case?

jantimon (changes requested):
I might be wrong but I don't think that `forbiddenValues` is needed and would be a big maintenance effort.

But maybe I don't understand it entirely - why would we need this?

Wouldn't it be the same problem for font, background, ...?

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
rvetere: nooow it's solved much better

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
rvetere: deleted again

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
rvetere: done

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
rvetere: done

packages/next-yak/loaders/__tests__/tsloader.test.ts
rvetere: solved

packages/next-yak/loaders/babel-yak-plugin.cjs
rvetere: done

packages/next-yak/loaders/babel-yak-plugin.cjs
rvetere: yes will think about... :) suggestions? :)

packages/next-yak/loaders/lib/extractCssUnit.cjs
rvetere: not there anymore

packages/next-yak/loaders/babel-yak-plugin.cjs
jantimon: not really maybe @Mad-Kat ?
if we can't find a better one `handle` is good enough

rvetere: "appendCssUnitsInExpression" evtl.?
"supportCssUnitInExpression"

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,61 @@
...
+          t.stringLiteral(cssUnit)
+        );
+
+        const newArrowFunction = t.arrowFunctionExpression(
+          expression.params,
+          newBody
+        );
+        return newArrowFunction;
+      }
+    }
```
Mad-Kat, line 46: Maybe it makes sense to move this statement outside of the `if` above. This would reduce nesting and makes it clearer that this is a separate case from the others

packages/next-yak/loaders/babel-yak-plugin.cjs
Mad-Kat: `replace` or `add`? But `append` would also work

Mad-Kat (approved):

packages/next-yak/loaders/babel-yak-plugin.cjs
Mad-Kat: `replace` or `add`? But `append` would also work

jantimon: maybe `moveCssUnitIntoExpression`?

jantimon: I like `append` most as it means add it to the end

`appendCssUnitToExpressionValue` ? 🤣

packages/next-yak/loaders/lib/extractCssUnit.cjs
```diff
@@ -0,0 +1,16 @@
...
+ *
+ * ```js
+ *   // in the AST this code would have two quasis: "margin: " and "px 0 14px;"
+ *   css`margin: ${size}px 0 14rem;` 
+ * ```
+ *
+ * @param {string} quasi
+ */
+function extractCssUnit(quasi) {
+  const match = quasi.match(/\s|([a-z%]+)(?=;|\s|\))/);
```
jantimon, line 12: We don't want to find any css unit but only leading css units as only leading units would follow after an expression.

e.g.:
```js
const extractLeadingCssUnit = (code) => code.match(/^([a-z]+|%)/i)?.[0]
```

Maybe now that it is really only `.match(/^([a-z]+|%)/i)?.[0]` we can keep it in the main file and don't have to create a dedicated file for a simple regexp

packages/next-yak/loaders/cssloader.cjs
```diff
@@ -234,11 +239,22 @@ module.exports = async function cssLoader(source) {
...
         const expression = path.node.quasi.expressions[i];
         // loop over all quasis belonging to the same css block
         const type = quasiTypes[i];
+
+        let code = unEscapeCssCode(quasi.value.raw);
+        if (removeCssUnit) {
+          // a quasi might start with a css unit e.g. css`margin: ${value}px 0;`
+          // the css unit will be part of the typescript runtime code and must
+          // must be removed from the css code as `var(--foo)px` would be invalid css
+          code = code.replace(/^([a-z]+|%)/, "");
```
jantimon, line 254: `PX` is valid css - I hope no one writes code like this but still we should add support and keep the regexp equal to the ts-loader

```suggestion
          code = code.replace(/^([a-z]+|%)/i, "");
```

jantimon (changes requested):

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
```diff
@@ -0,0 +1,61 @@
...
+    const binaryExpression = t.binaryExpression(
+      "+",
+      expression,
+      cssUnitLiteral
+    );
+    return binaryExpression;
+  }
+
+  // No unit found
+  return expression;
```
jantimon, line 58: the comment looks wrong to me - which cases would come here - should we throw an error?

packages/next-yak/loaders/babel-yak-plugin.cjs
rvetere: Ich übernehm mal `appendCssUnitToExpressionValue`, gefällt mir :)

packages/next-yak/loaders/lib/handleCssUnitInExpression.cjs
rvetere: i agree - i buildet it like this because first i was hoping i could re-use the "else" case of the arrow-function handling for all other cases, but this is not true anymore... so you're change makes totally sense!

jantimon (approved):

PR #66, opened 2024-02-09
add square-bracket-support

jantimon:
we got a request to support numeric indexes for yak files e.g.:

```ts
import { spacings } from "./constants.yak";

const Button = styled.button`
   margin: ${spacings[0]}
`
```

this pr uses recursions to allow any combination of computed and uncomputed member expressions like for example:

`a.usage["of"][5].member.expressions`

### DIFF
--- a/packages/next-yak/loaders/lib/replaceQuasiExpressionTokens.cjs
+++ b/packages/next-yak/loaders/lib/replaceQuasiExpressionTokens.cjs
@@ -27,41 +27,21 @@
 module.exports = function replaceTokensInQuasiExpressions(quasi, replacer, t) {
   for (let i = 0; i < quasi.expressions.length; i++) {
     const expression = quasi.expressions[i];
-    // replace direct identifiers e.g. ${query}
-    if (t.isIdentifier(expression)) {
-      const replacement = replacer(expression.name);
-      if (replacement === false) {
-        continue;
-      }
-      replaceExpressionAndMergeQuasis(quasi, i, replacement);
-      i--;
+    // find the value to replace the expression with
+    const replacement = getReplacement(expression, replacer, t);
+    if (replacement === false) {
+      continue;
     }
-    // replace member expressions e.g. ${query.xs}
-    // replace deeply nested member expressions e.g. ${query.xs.min}
-    else if (
-      t.isMemberExpression(expression) &&
-      t.isIdentifier(expression.object)
-    ) {
-      /** @type {any} */
-      let replacement = replacer(expression.object.name);
-      if (replacement === false) {
-        continue;
-      }
-      /** @type {import("@babel/types").Expression} */
-      let object = expression;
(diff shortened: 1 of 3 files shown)


packages/next-yak/loaders/lib/replaceQuasiExpressionTokens.cjs
```diff
@@ -27,41 +27,21 @@
...
+    const replacementValue = getReplacementValueForExpression(
+      expression,
+      replacement,
+      t
+    );
+    if (replacementValue === false) {
+      continue;
     }
+    replaceExpressionAndMergeQuasis(quasi, i, replacementValue);
+    i--;
```
Mad-Kat: I know that this change wasn't made during the lifetime of the branch, but I think it would be helpful to have a comment to indicate why we need to `i--`. Or maybe we could move this expression inside of `replaceExpressionAndMergeQuasis`. WDYT?

packages/next-yak/loaders/lib/replaceQuasiExpressionTokens.cjs
```diff
@@ -85,3 +65,98 @@ function replaceExpressionAndMergeQuasis(quasi, expressionIndex, replacement) {
...
+
+/**
+ * Find the replacement for the expression
+ *
+ * searches for:
+ *   - `x` -> x
+ *   - `x.y` -> x
+ *   - `x[0]` -> x
+ *   - `x()` -> x
+ *   - `x.y()` -> x
```
Mad-Kat, line 75: Do we support functions in yak files? Empty functions would be possible, but we don't check if they are empty here, or do I miss something?

packages/next-yak/loaders/lib/replaceQuasiExpressionTokens.cjs
```diff
@@ -27,41 +27,19 @@
...
+    // if it is a nested value, find the value of the expression
+    // e.g. x.y.z -> find the value of z
+    const replacementValue = replacement && getReplacementValueForExpression(
+      expression,
+      replacement,
+      t
+    );
+    if (replacementValue !== false) {
+      replaceExpressionAndMergeQuasis(quasi, i, replacementValue);
       i--;
```
Mad-Kat, line 41: I know this wasn't change, but what do you think about adding a comment why we have to do `i--`?

Mad-Kat (approved):

packages/next-yak/loaders/lib/replaceQuasiExpressionTokens.cjs
jantimon: this function would already support functions and return `x` 

however the second part `getReplacementValueForExpression` would not allow functions for now as we would have to think about what should be allowed e.g.

only `x()`
or also `x(1)`
or also `x(y)` if `y` is also from a `.yak` file

Mad-Kat (approved):

PR #68, opened 2024-02-14
Add :global() to inline css classes automatically

### LINKED ISSUE #53: automatically inject `:global`
next-yak automatically uses css modules behind the scenes and in nextjs css modules are local by default.
this means that selectors e.g. `html.dark-mode` will be scoped automatically to the component

for example

```tsx
const StyledDiv = styled.div`
  background-color: #f0f0f0;
  html.dark-mode & {
    background-color: #000;
  }
`;
```

becomes

```tsx
.StyledDiv-be2f45 {
  background-color: #f0f0f0;
  html.dark-mode-a4e345 & {
    background-color: #000;
  }
}
```

so although the user wrote `html.dark-mode` it would NOT match.

as this is unintuitive for most users and therefore next-yak should automatically inject `:global`

rvetere:
fixes #53 
Screenshot 2024-02-14 at 10 06 34

### DIFF
--- a/packages/next-yak/loaders/__tests__/cssloader.test.ts
+++ b/packages/next-yak/loaders/__tests__/cssloader.test.ts
@@ -657,11 +657,17 @@ const headline = css\`
         `
 import { css } from "next-yak";
 
+
 const headline = css\`
   color: red;
   .dark-mode {
     color: black;
   }
+  .my-fancy-class,
+  .my-button {
+    color: blue;
+  }
+  letter-spacing: 0.05em;
 \``
       )
     ).toMatchInlineSnapshot(`
@@ -670,6 +676,11 @@ const headline = css\`
         :global(dark-mode) {
           color: black;
         }
+        :global(my-fancy-class),
+        :global(my-button) {
+          color: blue;
+        }
+        letter-spacing: 0.05em;
       }"
     `);
   });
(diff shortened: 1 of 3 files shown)


packages/next-yak/loaders/__tests__/cssloader.test.ts
```diff
@@ -649,4 +649,40 @@ const headline = css\`
...
+    color: blue;
+  }
+  transition: all 0.3s ease-in-out;
+  letter-spacing: 0.05em;
+\``
+      )
+    ).toMatchInlineSnapshot(`
+      ".headline_0 {
+        color: red;
+       :global(dark-mode) {
```
Mad-Kat, line 676: this should be `:global(.dark-mode)`

Mad-Kat (commented):
Thank you for the PR :)

Could you add some more tests for the following cases:
```jsx
import { styled, keyframes } from "next-yak";

const Link = styled.a\`
  color: palevioletred;
\`

const Icon = styled.svg\`
  \${Link} & {
    color: red;
  }
\`
```

```jsx
import { styled, keyframes } from "next-yak";

const Component = styled.a\`
  .any-class:hover {
    color: red;
  }
\`;
```

```jsx
import { styled, keyframes } from "next-yak";

const Component = styled.a\`
  :where(.any-class) {
    color: red;
  }
\`;
```

Mad-Kat (changes requested):
I think it doesn't handle all cases yet, but feel free to change and rerequest a review once my points are fixed

jantimon:
I'll close this for now because of the conflicts - feel free to reopen :)

PR #71, opened 2024-02-21
prevent dollars to be passed to another function

jantimon:
there was a bug which caused invalid prop forwarding

this pr adds the following test and ensures that it passes:

```ts
it("should filter out properties starting with $ when passing to custom", () => {
  let forwardedProps = null;
  const Component = ({ className, style, ...props}) => {
    forwardedProps = props;
    return null;
  };
  const StyledComponent = styled(Component)``;
  const { container } = render(<StyledComponent $forwardedProp="notForwarded" />);

  expect(forwardedProps).toEqual({});
});
```


rvetere (approved):
looks good to me

packages/next-yak/runtime/__tests__/attrs.test.tsx
```diff
@@ -530,11 +517,8 @@ it("should pass theme if theme is overwritten", () => {
     ).toJSON(),
   ).toMatchInlineSnapshot(`
     <pre
-      $__attrs={true}
       className=""
       style={{}}
-    >
-      {"color":"blue"}
-    </pre>
+    />
```
Mad-Kat, line 522: There should be the theme prop passed, shouldn't it?

packages/next-yak/runtime/__tests__/styled.test.tsx
```diff
@@ -302,9 +329,7 @@ it("should not remove theme if theme is passed to wrapped element", () => {
     <div>
       <pre
         class="test-wrapper test"
-      >
-        {"anything":"test"}
-      </pre>
+      />
```
Mad-Kat, line 332: Same as above

Mad-Kat (changes requested):
Generally the change looks good and I don't know where the -24% execution time comes in, but I like it 😄 

I think the approach of not forwarding the props if it's not a yak-component is generally good, but we have to think about the case with the theme. 

What do you think?

packages/next-yak/runtime/__tests__/styled.test.tsx
```diff
@@ -302,9 +329,7 @@ it("should not remove theme if theme is passed to wrapped element", () => {
     <div>
       <pre
         class="test-wrapper test"
-      >
```
jantimon, line 305: this is probably wrong

jantimon:
> Generally the change looks good and I don't know where the -24% execution time comes in, but I like it 😄
> 
> I think the approach of not forwarding the props if it's not a yak-component is generally good, but we have to think about the case with the theme.
> 
> What do you think?

Maybe I found a fix - lets see if the pipeline is happy

jantimon:
The performance gain is that it now finally detects nested yak components and calls them directly without telling react 🤣

Mad-Kat (approved):
Good job. Really nice change

PR #81, opened 2024-03-16
Moved tsup configurations to config file

Mad-Kat:
Moved configurations to a new `tsup` configuration file.

This approach is more readable and maintainable than the current package scripts.


jantimon:
what is the idea behind this pr?

wouldn't that change the tsup config from explicit  to implicit and remove documentation (why we need it each step)?

Mad-Kat:
> what is the idea behind this pr?
> 
> wouldn't that change the tsup config from explicit to implicit and remove documentation (why we need it each step)?

I was working on a TypeScript & ESM version of our loaders and I wasn't too sure if I should add another build command for that. I found this solution and thought it would be more expressive. But you're right. In the previous version of the PR we would loose some explicitness, as we don't know why we splitted those commands explicitly. 

I've added some comments that should bring back this explicitness with the added benefit that the config is now fully typed and you can read up on options for specific build. 

What do you think?

jantimon (approved):
I guess we could extend the comments even more but this is already good enough as it keeps the same information we had before 👍

jantimon (approved):

PR #82, opened 2024-03-17
Typescript loaders

Mad-Kat:
This PR adds the ability to import loaders as ESM & CJS. 

As there is already a build step for all the other code, adding a build step for loaders doesn't add much overhead.

All built code will be generated in `/dist/*` to have a better distinction between built code and code one can manually change.


jantimon:
the idea behind not compiling nodejs code was easier debugging

but I guess the advantages with one dist folder and not to mix esm / cjs help are way stronger

PR #83, opened 2024-03-26
Typescript loaders

Mad-Kat:
This PR adds the ability to import loaders as ESM & CJS.

As there is already a build step for all the other code, adding a build step for loaders doesn't add much overhead.

All built code will be generated in /dist/* to have a better distinction between built code and code one can manually change.


packages/next-yak/withYak/index.ts
```diff
@@ -30,7 +30,7 @@ const addYak = (yakOptions: YakConfigOptions, nextConfig: NextConfig) => {
     });
     webpackConfig.module.rules.push({
       test: /\.yak\.module\.css$/,
-      loader: path.join(currentDir, "../loaders/cssloader.cjs"),
+      loader: path.join(currentDir, "../loaders/cssloader.js"),
```
Mad-Kat, line 33: I don't know if we need to separate CJS & ESM loaders depending on the `withYak` type

packages/next-yak/tsup.config.ts
```diff
@@ -89,8 +89,20 @@ export default defineConfig([
...
     target: "es2022",
-    outDir: "withYak",
+    outDir: "dist/withYak",
+  },
+  // loaders
+  {
+    entryPoints: ["loaders/cssloader.ts", "loaders/tsloader.ts"],
+    format: ["esm", "cjs"],
+    minify: false,
+    sourcemap: true,
```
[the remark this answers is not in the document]
jantimon: evtl für stacktraces

packages/next-yak/withYak/index.ts
jantimon: no - webpack 5 is happy with cjs files

packages/next-yak/dist/withYak/index.cjs
```diff
@@ -57,7 +57,7 @@ var addYak = (yakOptions, nextConfig) => {
     });
     webpackConfig.module.rules.push({
       test: /\.yak\.module\.css$/,
-      loader: import_node_path.default.join(currentDir, "../loaders/cssloader.cjs"),
+      loader: import_node_path.default.join(currentDir, "../loaders/cssloader.js"),
```
jantimon, line 60: I am not sure if that breaks cjs support - did you try it in a cjs app?

jantimon (approved):

packages/next-yak/withYak/index.ts
Mad-Kat: [As it turns out ](https://stackoverflow.com/a/76556992) Webpack > 5.80.0 can use both CJS and ESM
I will change it to `.cjs` for maximal compatibility

jantimon (approved):

PR #88, opened 2024-04-17
Adds a playground

### LINKED ISSUE #59: Add playground
Add the possibility to see what `next-yak` does when you use it in your next.js application.

Next-yak grew to a size so that it's not always manageable to fit every aspect of it in my head.

This would help to quickly discuss new feature ideas, to verify certain problems and to some extends help users to understand what's happening.

Mad-Kat:
WIP

closes #59 

Still todo/explore:
- ~~Add syntax highlighting for `styled-components` syntax~~  done
- ~~Integrate it in our docs package~~ thanks @jantimon 
- ~~Add the possibility to switch themes~~ done

Nice to have:
- Optimize it to use our babel plugin directly so that we could use `@babel/standalone` and don't need to rely on bundling `@babel/core` and polyfill node API's, also the bundle would get smaller
- Add the possibility to switch `next-yak` versions & generate static links (for error reproduction)
- Have a look into https://sandpack.codesandbox.io/


packages/playground/src/useVocsTheme.ts
```diff
@@ -0,0 +1,32 @@
...
+      const storedTheme = localStorage.getItem("vocs.theme");
+      if (storedTheme) return storedTheme;
+    }
+    return window.matchMedia("(prefers-color-scheme: dark)").matches
+      ? "dark"
+      : "light";
+  });
+
+  useEffect(() => {
+    if (!document) return;
```
jantimon, line 17: in an effect `document` is never empty

packages/playground/src/useVocsTheme.ts
```diff
@@ -0,0 +1,32 @@
...
+    const htmlElement = document.getElementsByTagName("html")?.[0];
```
jantimon, line 18: `const htmlElement = document.documentElement`

packages/playground/src/useVocsTheme.ts
```diff
@@ -0,0 +1,32 @@
...
+    if (htmlElement) {
```
jantimon, line 19: document.documentElement is always present

packages/playground/src/useVocsTheme.ts
```diff
@@ -0,0 +1,32 @@
...
+      const observer = new MutationObserver((e) => {
+        if (e[0].attributeName === "class") {
```
jantimon, line 21: maybe deconstructure e?

jantimon (dismissed):

jantimon (approved):

PR #104, opened 2024-06-24
Fix composition types

### LINKED ISSUE #72: TypeScript error when trying to style component inside a styled component
**To Reproduce**
https://stackblitz.com/edit/stackblitz-starters-6risew?file=app%2Fpage.tsx
Here's a minimal code snippet to reproduce the issue:

```tsx

const Child = styled.div<{ $hasError: boolean }>`
  background: ${({ $hasError }) => ($hasError ? 'red' : 'green')}
`;

const Parent = styled.div`
  ${Child} {
    margin-top: 16px;
  }
`;

export default () => (
  <Parent>
    {/* ^^ Property '$hasError' is missing in type '{ children: Element[]; }' but required in type '{ $hasError: boolean; }'. */}
    <div>foo</div>
    <Child $hasError>bar</Child>
  </Parent>
);

### PR DESCRIPTION
**Pull request by Mad-Kat**
Fixes #72 

Use `NoInfer<T>` in order to be explicit for the tagged template literal types and not inferring them from the functions that are used for dynamic props or child components.

Added bonus: The destructuring of the props in the tagged template literal gives you now type hints, as TypeScript doesn't want to infer it dynamically and the type is known when you start with the literal.

### DIFF
--- a/packages/next-yak/package.json
+++ b/packages/next-yak/package.json
@@ -78,7 +78,7 @@
     "jsdom": "^22.1.0",
     "react": "^18.2.0",
     "react-test-renderer": "^18.2.0",
-    "typescript": "5.1.6",
+    "typescript": "5.4.5",
     "vite": "^4.4.9",
     "vitest": "0.34.5",
     "tsup": "7.2.0"
--- a/packages/next-yak/runtime/__tests__/typeTest.tsx
+++ b/packages/next-yak/runtime/__tests__/typeTest.tsx
@@ -190,11 +190,15 @@ const CompositionOverridingAndMergingTest = () => {
       ${Child} {
         color: blue;
       }
-      color: ${({ $colorMe }) => ($colorMe ? `blue` : `red`)};
+      color: ${
+        // @ts-expect-error - should not allow props which have not been defined (like styled-components)
+        ({ $colorMe }) => ($colorMe ? `blue` : `red`)
+      };
     `;
 
     // Overriding with Child and Mixin should work
     const _ = (
+      // @ts-expect-error - should not allow props which have not been defined (like styled-components)
       <Parent $colorMe>
         <Child $child />
       </Parent>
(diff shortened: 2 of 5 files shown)

### CODE REVIEW

**Comment by jantimon** [comment]
Would this be valid in styled components?

```tsx
  const Child = styled.div<{ $child: boolean }>`
      ${({ $child }) =>
        $child &&
        css`
          color: red;
        `}
    `;

    const Parent = styled.div`
      ${Child} {
        color: blue;
      }
      color: ${({ $colorMe }) => $colorMe ? `blue` : `red`};
    `;
```

It seems to fail for us (see broken pipeline error):
`Property '$colorMe' does not exist on type 'ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { $child: boolean; } & { theme: YakTheme; }'.ts(2339)`

jantimon (changes requested):
Would this be valid in styled components?

```tsx
  const Child = styled.div<{ $child: boolean }>`
      ${({ $child }) =>
        $child &&
        css`
          color: red;
        `}
    `;

    const Parent = styled.div`
      ${Child} {
        color: blue;
      }
      color: ${({ $colorMe }) => $colorMe ? `blue` : `red`};
    `;
```

It seems to fail for us (see broken pipeline error):
`Property '$colorMe' does not exist on type 'ClassAttributes<HTMLDivElement> & HTMLAttributes<HTMLDivElement> & { $child: boolean; } & { theme: YakTheme; }'.ts(2339)`

jantimon:
Just checked and styled-component seems to behave the same:
https://stackblitz.com/edit/vitejs-vite-vttcyb?file=src%2Fmain.tsx,src%2FApp.tsx&terminal=dev

I adjusted the tests and added @ts-expect-error

Mad-Kat:
Yeah I think you need to write it like this:

```
const Parent = styled.div<{ $colorMe: boolean }>`
      ${Child} {
        color: blue;
      }
      color: ${({ $colorMe }) => $colorMe ? `blue` : `red`};
    `;
```

you have to add the type parameter in order to say that `$colorMe` exists. It doesn't automatically infer it from the functions you provide.

jantimon:
released as 0.2.2

PR #105, opened 2024-06-29
add bundler integration tests for webpack

jantimon:
This PR introduces a new private package for the next-yak monorepository
It is called `webpack-tests` and is meant to test the integration into webpack for our bundling process

Key additions and improvements:

1. New `packages/webpack-tests` directory with comprehensive Webpack integration tests
2. Custom compilation function that simulates Next.js Webpack configuration and build process
3. In-memory file system setup for efficient and isolated testing
4. Tests to verify correct CSS extraction from TypeScript files using our custom loaders
5. Integration of `next-yak/withYak` into the build process (allows to test different configs)
6. Will allow us to run babel and swc tests with the same unit tests

The new test suite runs as part of our CI pipeline, adding an extra layer of confidence to our build and release process.

### DIFF
--- a/packages/next-yak/package.json
+++ b/packages/next-yak/package.json
@@ -78,7 +78,6 @@
     "jsdom": "^22.1.0",
     "react": "^18.2.0",
     "react-test-renderer": "^18.2.0",
-    "next": "14.0.0",
     "typescript": "5.5.2",
     "vite": "^4.4.9",
     "vitest": "0.34.5",
--- a/packages/next-yak/package.json
+++ b/packages/next-yak/package.json
@@ -78,6 +78,7 @@
     "jsdom": "^22.1.0",
     "react": "^18.2.0",
     "react-test-renderer": "^18.2.0",
+    "next": "14.0.0",
     "typescript": "5.5.2",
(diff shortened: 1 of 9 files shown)


packages/next-yak/package.json
```diff
@@ -78,6 +78,7 @@
     "jsdom": "^22.1.0",
     "react": "^18.2.0",
     "react-test-renderer": "^18.2.0",
+    "next": "14.0.0",
```
Mad-Kat, line 81: Do we need this dependency for `next-yak`?

Mad-Kat (dismissed):

packages/next-yak/package.json
jantimon: you are correct that should not be part of this pr

I tested the new bundler integration tests with the built cross-file selector branch version of next-yak.
And the cross-file selector needs next to be installed because of https://github.com/jantimon/next-yak/blob/a7cbc5ba439d6abab04da6357fefa3929b36535d/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts#L7

I'll remove it from this branch

PR #107, opened 2024-06-30
reuse compilation result of ts-loader

jantimon:
right now the special import e.g. `./page.yak.module.css!=!./page?./page.yak.module.css` is executed by the css-loader directly from it's original source

this means that the typescript transformation has to be run again for the entire code although it was just transformed by the ts-loader

to workaround this issue the compilation result of `./page` could be reused for `./page.yak.module.css!=!./page?./page.yak.module.css`

such a workaround could also help us for `swc` as executing swc from the css-loader with the very same options might be tricky


packages/next-yak/loaders/cssloader.ts
```diff
@@ -9,27 +9,13 @@ import { resolveCrossFileSelectors } from "./lib/resolveCrossFileSelectors.js";
...
-        }
-        const css = extractCss(String(code));
-        return resolveCrossFileSelectors(this, css).then(
-          (result) => callback(null, result),
-          callback,
-        );
-      },
-    },
-    source,
+  const typescriptCode = getTsLoaderResultFromCache(this);
```
jantimon, line 14: Unfortunately this breaks as soon as an experimental cross file dependency changes.

The reason is that `resolveCrossFileSelectors` adds additional dependencies for this loader (the cssloader) which will reexecute the cssloader once those dependencies change. E.g. after changing a constant in another file.

In that edgecase the tsloader didn't run (because none of its dependencies changed) and the cssloader crashes the build

jantimon:
I am no longer able to build the playground locally 🤷‍♂️

@Mad-Kat could you please take a look?

package.json
```diff
@@ -16,8 +16,7 @@
     "prettier": "pnpm run --recursive --if-present prettier"
   },
   "engines": {
-    "node": ">=20",
-    "pnpm": "8"
```
Mad-Kat, line 20: I'm not so happy with this constraint, as I have `9.5` locally. Is this the reason the deployment fails currently?

Mad-Kat (dismissed):

Mad-Kat (approved):

package.json
jantimon: it has to be in sync with the ci config in `.github`

the build runs again - next canary points now to 15.x

Mad-Kat (approved):

PR #108, opened 2024-06-30
Feature/fix stackblitz

jantimon:
follow up for #108

once cssloader does not import tsloader we can switch from mjs to cjs

this might fix a incompatibility with stackblitz (https://github.com/stackblitz/webcontainer-core/issues/1478)

### DIFF
--- a/packages/next-yak/cssloader.js
+++ b/packages/next-yak/cssloader.js
@@ -1,0 +1,1 @@
+module.exports = require("./dist/loader/cssloader.cjs");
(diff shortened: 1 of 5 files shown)


packages/next-yak/cssloader.js
Mad-Kat, line 1: Why do we need this file? In `withYak` we use loader directly

packages/next-yak/tsloader.js
Mad-Kat, line 1: Why do we need this file? In `withYak` we use loader directly

Mad-Kat (changes requested):
LGTM, but I'm not sure if the additional files are really needed

PR #111, opened 2024-07-05
Fix comments issue

kdy1:
Comments are attached to `BytePos`, so you should not replace the span of the node with `DUMMY_SP` afterwards


jantimon:
@kdy1 thank you so much

jantimon:
It really works now with the fixture test!! 

shot-Hk0CbnZG@2x

Issue #56, opened 2024-02-01
add support for vitest / jest


jantimon:
find a way how next-yak does not prevent users from writing ui components tests

jantimon:
#### Possible solution

We could split the code so that all APIs which get transpiled by the Babel plugin (like `styled`, `keyframes`, `css`, etc.) to a new `next-yak/internal` module.   
Here's how it would work:

1. Our Babel plugin would automatically replace imports like `import { styled } from "next-yak"` with `import { styled } from "next-yak/internal"`
2. We'd keep a test-friendly version of these APIs in the main `next-yak` export

#### Benefits

For our users:
- Vitest/Jest would "just work" out of the box
- No need for extra configuration in your test setup
- Potentially better / optimized snapshot tests without hashes

For us maintainers:
- Cleaner typings: Currently, our type definitions are a bit messy. The `import { styled } from "next-yak"` statement has different meanings before and after compilation. By separating into two different exports, we can properly type both scenarios:
  - Pre-compilation: The external API that users interact with
  - Post-compilation: The internal API that represents how the code actually runs
- Improved unit testing:
  - We could structure our tests with separate `input.tsx` (before compilation) and `output.tsx` (after compilation) files
  - `input.tsx` would import from "next-yak" (simulating user code)
  - `output.tsx` would import from "next-yak/internal" (representing compiled code)
  - This setup would give us proper type checking for our compilation results in `output.tsx`
- With a clear separation, we can better document what's intended for public use vs. what's for internal operations, potentially reducing confusion for contributors

#### Potential Downsides

- Bundle size reporting on sites like Bundlephobia might not be accurate anymore
- Users who don't set up the compilation steps correctly might see the styles "working" (no errors) but not actually being applied

PR #117, opened 2024-07-12
create next-yak/internal

jantimon:
this is the very first step for #56 

it creates the structure and the adjusts the babel plugin to split the internal and external usage

by now it should work exactly the same (this means pr does not include any improvements for testing)

## Summary by CodeRabbit

- **New Features**
  - Introduced internal module configuration for better separation between public API and internal logic.
  - Added mock implementations for CSS-related functionality to support testing.

- **Refactor**
  - Adjusted import paths to use internal modules for cleaner code organization and improved typings.

- **Documentation**
  - Enhanced comments for clarity on usage in tests versus production.


Mad-Kat (dismissed):
LGTM
But just something that came to my mind:
Maybe it makes sense to change the import in `jsx-runtime.ts` as well. Wdyt?

jantimon:
> LGTM But just something that came to my mind: Maybe it makes sense to change the import in `jsx-runtime.ts` as well. Wdyt?

Yes you are right - the mocks will be the public facing typing and jsx-runtime.ts too.. therefore jsx-runtime must not export the internal types

Mad-Kat (approved):

PR #118, opened 2024-07-12
create keyframes mock

jantimon:
this is the `keyframes` mock implementation for #56 

during jest/vite it's a noop because there is no css extraction compile step

## Summary by CodeRabbit

- **New Features**
  - Added a mock implementation for the `keyframes` function to support testing of CSS keyframe animations.


Mad-Kat:
Doesn't this mean that the dynamic tagged templates aren't correctly typed, as mocks is exposed to the outside world?

jantimon:
those mocks export the public types - so they must be shipped with the correct public typings 

did I somehow break those types?

Mad-Kat:
Nono. I wasn't sure if you changed the types as the type signalizes that you can't use any interpolations. But I checked now and it's basically the same as the previous type so that should work :)

Mad-Kat (approved):

PR #119, opened 2024-07-12
css literal mock

jantimon:
this is the `css` mock implementation for #56

during jest/vite it's not returning any class names because there is no css extraction compile step

## Summary by CodeRabbit

- **New Features**
	- Enhanced the `css` function to support dynamic CSS expressions, improving styling flexibility.


packages/next-yak/runtime/mocks/cssLiteral.ts
```diff
@@ -1 +1,103 @@
...
+    // compiled -> css((props: { active: boolean }) => props.active && css("yak31e4"))
+    if (typeof arg === "function") {
+      dynamicCssFunctions.push(arg as unknown as PropsToClassNameFn);
+    }
+  }
+  if (dynamicCssFunctions.length === 0) {
+    return {
+      className: "css-mixin",
+      style: undefined,
+    };
```
Mad-Kat, line 81: Why do we return something here if we don't have any interpolations and nothing if we have interpolations?

Mad-Kat (approved):

packages/next-yak/runtime/mocks/cssLiteral.ts
jantimon: good point - that doesn't look right

Mad-Kat (approved):

PR #125, opened 2024-07-14
enabling step debug logging for codspeed

jantimon:
Adrien gave us a great hint to find out why codspeed became flaky:

> By looking at this log from a run, it seems that now the node flags needed to make node's execution non-flaky are no longer present: https://github.com/jantimon/next-yak/actions/runs/9926270270/job/27419653757?pr=120#step:10:27
> To continue investigating, can you activate step debug logging:https://docs.github.com/en/actions/monitoring-and-troubleshooting-workflows/enabling-debug-logging#enabling-step-debug-logging, and create a PR?

## Summary by CodeRabbit

- **Chores**
  - Added `ACTIONS_STEP_DEBUG: true` environment variable in the benchmarks job to aid in debugging.
  - Updated the workflow file `codspeed.yml` to improve indentation of branch lists and paths.
  - Upgraded the action version for running benchmarks from `v1` to `v2` to ensure compatibility and access to the latest features.

### DIFF
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -18,6 +18,8 @@ on:
 jobs:
   benchmarks:
     runs-on: ubuntu-latest
+    env:
+      ACTIONS_STEP_DEBUG: true
     steps:
       - name: Checkout
         uses: actions/checkout@v3
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -15,10 +15,6 @@ on:
       - ".github/workflows/codspeed.yml"
   workflow_dispatch:
 
-# enable step debug logging
-env:
-  ACTIONS_STEP_DEBUG: true
-
 jobs:
   benchmarks:
     runs-on: ubuntu-latest
--- a/.github/workflows/codspeed.yml
+++ b/.github/workflows/codspeed.yml
@@ -18,8 +18,6 @@ on:
 jobs:
   benchmarks:
     runs-on: ubuntu-latest
-    env:
-      ACTIONS_STEP_DEBUG: true
     steps:
       - name: Checkout
         uses: actions/checkout@v3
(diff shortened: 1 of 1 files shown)


adriencaccia:
Hey @jantimon, thanks for opening the PR!
In order for the step debug logging to work, you need to set the `ACTIONS_STEP_DEBUG` at the repository level, either in the actions variables or secrets.

Can you also bump the CodSpeed action to v2? There is no breaking changes, and it will allow me to have more information on my end

jantimon:
Added a new secret:
shot-9aLnKo4X@2x

.github/workflows/codspeed.yml
```diff
@@ -46,9 +48,9 @@ jobs:
       - name: Build Yak
         run: pnpm run build
       - name: Build benchmark suite
-        run: cd packages/benchmark && pnpm run codspeed 
+        run: cd packages/benchmark && pnpm run codspeed
       - name: Run benchmarks
-        uses: CodSpeedHQ/action@v1
+        uses: CodSpeedHQ/action@v2
```
adriencaccia, line 53: Can you also add the `CODSPEED_DEBUG: true` env variable to the codspeed step please?
It seems there is a problem with the node introspection.

jantimon: done 👍

.github/workflows/codspeed.yml
```diff
@@ -46,9 +46,9 @@ jobs:
...
```
adriencaccia, line 51: I do not see the debug logs in the workflow 😖 
Can you add the env variable like so?

```suggestion
        uses: CodSpeedHQ/action@v2
        env:
          CODSPEED_DEBUG: true
```

Mad-Kat: I've added it. Thanks for looking into it!

jantimon:
@adriencaccia wow really cool thank you :)

can we disable step debugging again?

adriencaccia:
> can we disable step debugging again?

Yes, sure!

PR #134, opened 2024-07-19
Add shareable playground links

Mad-Kat:
This adds the ability of sharing a specific playground state with a link.

It's inspired by the action in [TS Playground](https://www.typescriptlang.org/play/).

This adds the `ctrl/cmd + s` action to get the sharable link in the current clipboard or when clicking outside of the editor. Additionally this is also added to the context menu when right clicking in the editor:

image


jantimon (approved):

Mad-Kat:
The problem is that Codspeed doesn't run on this PR, because I didn't touch any performance relevant code but it's still required to run in order to merge the PR. Is it possible to make Codspeed only required when you touch the performance relevant code?

PR #141, opened 2024-07-23
ignore empty class names

jantimon:
this improvement helps with snapshot results

### DIFF
--- a/packages/next-yak/runtime/__tests__/attrs.test.tsx
+++ b/packages/next-yak/runtime/__tests__/attrs.test.tsx
@@ -14,7 +14,6 @@ it("works fine with an empty object", () => {
   const Comp = styled.div.attrs({})``;
   expect(TestRenderer.create(<Comp />).toJSON()).toMatchInlineSnapshot(`
     <div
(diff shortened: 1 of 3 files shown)


packages/next-yak/runtime/__tests__/attrs.test.tsx
```diff
@@ -464,18 +443,16 @@ it("should have optional attrs props as component interface", () => {
...
 
   expect(TestRenderer.create(<Comp $primary />).toJSON())
     .toMatchInlineSnapshot(`
-    <h1
-      className=""
-      style={{}}
-    />
-  `);
+      <h1
+        style={{}}
```
Mad-Kat, line 453: Could we do the same with empty styles?

Mad-Kat (approved):

packages/next-yak/runtime/__tests__/attrs.test.tsx
jantimon: I believe this would be more expensive but especially for testing useful 👍

PR #144, opened 2024-07-26
rework resolving

### LINKED ISSUE #142: enhance CSS constant import resolution
#### Background

Our CSS-in-JS system allows for cross-file references within CSS expressions in TypeScript source files - like in this example:

```typescript
// theme.ts
export const colors = {
  primary: '#3498db',
  secondary: '#2ecc71'
};

// Button.tsx
import { colors } from './theme';

const Button = styled.button`
  background-color: ${colors.primary};
  &:hover {
    background-color: ${colors.secondary};
  }
`;

```

Currently, our CSS loader resolves these imports using two methods:

1. For normal imports (not  .yak.ts), it uses Webpack's resolve API to find the file, then uses Babel to look for exports and re-exports, recursively resolving until it finds the final exported value

2. For imports from .yak.ts files, it evaluates the entire file using Node (via Webpack's evaluation API)

#### Proposed Changes
1. Extend the recursive resolution process to look for and evaluate .yak.ts files at each step of the resolution, not just for the initial import
2. When resolving an import:
   a. First, attempt to resolve as a normal import (current behavior).
   b. If the normal resolution reexports a value from a .yak.ts file, evaluate it using Node (as we currently do for initial .yak.ts imports).
   d. Continue this process recursively until the final value is resolved.

#### Benefits
1. Consistency: Applies the same logic for .yak.ts files throughout the resolution process
2. Mono Repositories package.json exports don't have to include the .yak extensions and can mix static and evaluated values

jantimon:
1. Improved Caching Strategy:
   - Cache keys now use resolved paths for more reliable caching across different import scenarios.

2. Recursive Resolution Improvements:
   - As soon as a reexport is form a `.yak` file the file is evaluated

3. Support for Star Exports:
   - Added ability to follow 'export * from' statements (when it's **the only star export** and the identifier wasn't found elsewhere)
   
4. Enhanced Error Handling:
   - Improved error messages throughout the code, providing more context about where errors occur

Fixes #142


packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
```diff
@@ -7,412 +7,433 @@ import { getCssModuleLocalIdent } from "next/dist/build/webpack/config/blocks/cs
...
 
-type RecursiveRecord = {
-  [key: string]: RecursiveRecord | number | string;
-};
+async function parseFile(
+  loader: LoaderContext<{}>,
+  filePath: string,
+): Promise<ParsedFile> {
+  const isYak = filePath.endsWith(".yak.ts") || filePath.endsWith(".yak.tsx");
+  const isTSX = filePath.endsWith(".tsx");
```
Mad-Kat, line 169: What about `.js` and `.jsx`?

Mad-Kat (dismissed):

packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
jantimon: good point 👍

PR #149, opened 2024-08-02
Feature/better mixins

### LINKED ISSUE #146: redefining css mixins
#### Summary

Introducing a new ``const colorMixin = atom.css`color:red`;`` API and redefining the behavior of the existing `css` template literal in next-yak could solve the cross file css mixin complexity

#### Background

Currently, next-yak treats all `css` template literals as static, reusable atomic classes. While this approach works well for simple cases and promotes CSS reuse, it falls short in more complex scenarios, particularly when used within nested selectors or media queries

For example:

```typescript
const borderMixin = css`
  border: 1px solid black;
  border-radius: 999px;
`; // becomes .border-mixin { border: 1px solid black; border-radius: 999px; }

const Button = styled.button`
  &:hover {
    ${borderMixin}
  }
  color: red;
`; // becomes <button class="button border-mixin">...</button?
``` 

In this case, the current implementation cannot correctly apply the `borderMixin` within the `:hover` selector, as it's treated as a static atomic class

@Mad-Kat pointed out that this is a core problem where a lot of automated atomic utility css extraction get very complex

#### Proposed Solution

Instead of creating atomic css mixins automatically we make it an optional feature, so developers can explicitly create an atom and are aware of the limitations or just use the ordinary css api with a little bit more bundled css  code

So we would:

1. Introduce a new `atom.css` API for creating static, atomic CSS mixins 
2. Redefine the behavior of the existing `css` template literal to always inline its contents when used

##### 1. New `atom.css` API

The `atom.css` API will behave similarly to the current `css` implementation, creating a static, reusable atomic class:

```typescript
const borderMixin = atom.css`
  border: 1px solid black;
  border-radius: 999px;
`;
```

This will be extracted as a utility atom with a class (e.g., `border-mixin`) and can be reused across the codebase, reducing bundled CSS size.

##### 2. Redefined `css` Behavior

The `css` template literal will now inline its contents when used, allowing for proper nesting and selector combination:

```typescript
const borderMixin = css`
  border: 1px solid black;
  border-radius: 999px;
`;

const Button = styled.button`
  &:hover {
    ${borderMixin}
  }
  color: red;
`;
```

This change allows the `borderMixin` to be correctly applied within the `:hover` selector.

#### Cross File behaviour

This could also work for cross file mixins.  
Here is an example how the compilation would look like:

##### Source Code

```typescript
import { borderMixin } from "./mixin";
const Button = styled.button`
  &:hover {
    ${borderMixin}
  }
  color: red;
`;
```

##### Compiled Output

The above code would be compiled to:

```typescript
import { borderMixin } from "./mixin";
const Button =
/*YAK Extracted CSS:
.Button {
  &:hover { --yak-css-import: url("./mixin:borderMixin"); }
  color: red;
}*/
/*#__PURE__*/
styled.button(borderMixin, __styleYak.Button);
```

##### CSS Loader Processing

The CSS loader would then process the extracted CSS, replacing the special `--yak-css-import` property with the actual content of the `borderMixin`. For example, if `borderMixin` was defined as:

```typescript
export const borderMixin = css`
  border: 1px solid black;
  border-radius: 999px;
`;
```

The final CSS output would be:

```css
.Button {
  &:hover {
    border: 1px solid black;
    border-radius: 999px;
  }
  color: red;
}
```

As we keep the mixin reference also in the typescript code (`styled.button(borderMixin, __styleYak.Button);`) this approach could also work for dynamic mixins across files

Mad-Kat:
Closes #146 

shot-KENUsk9L@2x


jantimon:
we should probably e sure that top level css mixing (e.g. ``const foo  = css`..```

are only defined in const statements and/or think about object definitions like 

```
const typography = {
  h1: css`..`
}
```

packages/yak-swc/yak_swc/tests/fixture/cross-file-mixin-import/output.tsx
```diff
@@ -11,29 +11,38 @@ export const Button = /*YAK Extracted CSS:
...
   color: green;
 }
-*/ /*#__PURE__*/ styled.button(__styleYak.Button5, fonts.h1, fancy);
+*/ /*#__PURE__*/ styled.button(__styleYak.Button5, fonts.h1, fancy.mixins.specialEffect);
 export const Button6 = /*YAK Extracted CSS:
 .Button6 {
   &:hover {
     --yak-css-import: url("./constants.yak:yakMixin");
   }
+  --yak-css-import: url("./fancy:fancy:fancy:mixins:specialEffect")
```
jantimon, line 44: this looks wrong - it should be inside `&:hover`

packages/yak-swc/yak_swc/tests/fixture/mixing-with-tailwind-error/output.stderr
```diff
@@ -1,7 +0,0 @@
-  x Inside nested selectors you can only use css literals, constants or dynamic values
```
jantimon, line 1: this is a bug - we must not allow this

jantimon (approved):

PR #152, opened 2024-08-12
Enable tests that contain tsx

Mad-Kat:


jantimon:
we will hold this for now at it would break nextjs compatibility (which runs on a older swc core)

Issue #166, opened 2024-10-03
Keyframe animations defined after usage not resolved correctly


jantimon:
When keyframe animations are defined after their usage in styled components, particularly within nested css template literals in conditional statements, the plugin throws an "Unsupported template literal in css expression" error. This issue occurs even though the keyframes are defined in the same file.

## Minimal Reproducible Example

```typescript
import { css, keyframes, styled } from "next-yak";

const Card = styled.div<{ $flipped: boolean }>`
  position: relative;
  transform-style: preserve-3d;
  transform: rotateY(180deg);

  ${({ $flipped = false }) =>
    $flipped
      ? css`
          animation: ${cardFlip} 0.275s forwards linear;
        `
      : css`
          animation: ${cardUnflip} 0.275s forwards linear;
        `}
`;

// Keyframes defined after usage
const cardFlip = keyframes`
  0% {
    transform: rotateZ(0deg) rotateY(180deg);
  }
  50% {
    transform: rotateZ(-10deg) rotateY(90deg);
  }
  100% {
    transform: rotateZ(0deg) rotateY(0deg);
  }
`;

const cardUnflip = keyframes`
  0% {
    transform: rotateZ(0deg) rotateY(0deg);
  }
  50% {
    transform: rotateZ(-10deg) rotateY(90deg);
  }
  100% {
    transform: rotateZ(0deg) rotateY(180deg);
  }
`;
```

## Error Message

```
Unsupported "cardFlip" template literal in css expression - only css`` allowed
```

## Expected Behavior

The keyframe animations should be resolved and usable within the styled component, even when they are defined after their usage in the file.

The issue likely stems from the order of processing in the AST traversal.

jaggli:
It works, if the keyframe calls are before the Card definition, but it doesn't work if they're below (in the file) - so ordering seems to be an issue.

PR #167, opened 2024-10-07
fix keyframes order

jantimon:
This pull request addresses issue #166, where keyframe animations defined after their usage in styled components cause an "Unsupported template literal in css expression" error. The solution involves the following key changes:

- Introduced `ScopedVariableReference` to manage JavaScript variable references more effectively, ensuring that both simple and complex references (e.g., `foo` and `foo.bar`) are handled consistently
- Updated the handling of keyframe animations and nested CSS expressions to resolve references correctly, even when keyframes are defined after their usage

The solution is to link keyframe references with its declaration even before it was declared and reuse this name later during the actual declaration. This ensures that animations work seamlessly regardless of their order in the file

### DIFF
--- a/packages/yak-swc/yak_swc/src/utils/ast_helper.rs
+++ b/packages/yak-swc/yak_swc/src/utils/ast_helper.rs
@@ -92,18 +92,21 @@ pub fn create_member_prop_from_string(s: String) -> MemberProp {
 /// 2. Simple identifiers (e.g., `primaryColor`) -> Some((primaryColor#0, ["primaryColor"]))
 pub fn extract_ident_and_parts(expr: &Expr) -> Option<ScopedVariableReference> {
   match &expr {
-    Expr::Member(member) => member_expr_to_strings(member).map_or_else(
+    Expr::Member(member_expr) => member_expr_to_strings(member_expr).map_or_else(
       || {
         HANDLER.with(|handler| {
           handler
-            .struct_span_err(member.span, "Could not parse member expression")
+            .struct_span_err(member_expr.span, "Could not parse member expression")
             .emit();
         });
         None
       },
-      |member_exp_strings| {
-        let (root_ident, props) = member_exp_strings;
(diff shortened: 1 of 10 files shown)


jantimon:
This PR added support for keyframes in objects e.g.:

- ``const foo = { bar: keyframes`...` }``

The first version supported only identifiers so the following versions were ignored:

- ``const foo = { "bar": keyframes`...` }``
- ``const foo = { 123: keyframes`...` }``

So in this update I enhanced `visit_mut_object_lit` function to support more types of the PropName:
   - `Ident`: Uses the `sym` field directly
   - `Str`: Uses the `value` field
   - `Num`: Converts to string
   - `BigInt`: Converts to string
   - `Computed`: Skipped (cannot be statically analyzed)

I also added an additional test

packages/yak-swc/yak_swc/src/utils/ast_helper.rs
```diff
@@ -88,26 +90,35 @@ pub fn create_member_prop_from_string(s: String) -> MemberProp {
...
+      || {
+        HANDLER.with(|handler| {
+          handler
+            .struct_span_err(member.span, "Could not parse member expression")
+            .emit();
+        });
+        None
+      },
+      |member_exp_strings| {
+        let (root_ident, props) = member_exp_strings;
```
Tyderion, line 105: i don't think props is the best name here as it contains the rest of the member-expression path

jantimon: I looked into variables and changed most of them:

`expr` -> Keep as is
`member` -> `member_expr`
`member_exp_strings` -> `member_parts`
`root_ident` -> `base_ident`
`props` -> `member_chain`

Tyderion: i think that's way better 👍

Tyderion (approved):
lgtm

PR #169, opened 2024-10-09
Re-add possibility to add CSS prop

Mad-Kat:
With the move to the SWC plugin we lost the ability to transform css properties (#78).

This PR re-adds the possibility to use `next-yak` with css props

Still to discuss:
- How to handle errors? Should we silently fail and don't transform (if possible)


jantimon (approved):

jantimon:
Released in next-yak 2.3.0

PR #183, opened 2024-10-28
support element switching via `attrs({as: ... })`

### LINKED ISSUE #184: Allow limited polymorphic components through `attrs`
Currently, next-yak does not support [polymorphic components](https://styled-components.com/docs/api#as-polymorphic-prop) (switching elements via an `as` prop) like styled-components does.

This was an intentional decision for the several reasons:

1. **Semantic HTML & Accessibility**: When you create a component like `styled.button`, you're making a deliberate choice about its semantic meaning and accessibility behavior. HTML elements have built-in keyboard handling, focus management, and ARIA roles. Allowing arbitrary element switching via `<Button as="div">` can lead to semantically incorrect HTML structure and break critical accessibility features. This affects not just user experience but also SEO and assistive technology compatibility.

2. **Component Contract**: Styled components often include specific event handlers and props tied to their element type (like `disabled` for buttons). Unrestricted element switching can break these assumptions and create confusing behavior:

```tsx
// What happens to disabled and onClick when switched to div?
const Button = styled.button`
  &:disabled { opacity: 0.5 }
`
```

3. **TypeScript Complexity**: Supporting unrestricted polymorphic components requires extremely complex type definitions that need to handle all possible HTML elements while maintaining proper prop typing. This complexity makes the types harder to maintain, debug, and understand for users

However, we stumbled across some legitimate use cases where controlled element switching is valuable, like for example consistent button/link stylings

Rather than implementing unrestricted polymorphic components, we could use the `attrs` API for element switching at component definition time:

```tsx

// Base button component
const Button = styled.button`
  padding: 8px 16px;
  background: blue;
  color: white;
  border-radius: 4px;
  /* ... more styles ... */
`

// Need to reuse button styles for a link
const LinkButton = styled(Button).attrs({
  as: 'a'
})``;

// Usage
<Button onClick={() => {}}>Normal Button</Button>
<LinkButton href="/about">Link that looks like a button</LinkButton>
```

jantimon:
this PR adds runtime support for switching HTML elements through the `attrs` API:

```ts
const Button = styled.button`/* ... */`
const LinkButton = styled(Button).attrs({ as: 'a' })``
```

@Mad-Kat The runtime code is in place (in `runtime/styled.tsx`). Could you please add TypeScript types to make `as` only available in `attrs` objects but not as a prop on styled components?

fixes #184

### DIFF
--- a/packages/next-yak/runtime/styled.tsx
+++ b/packages/next-yak/runtime/styled.tsx
@@ -214,6 +214,10 @@ const yakStyled = <
         return parentYakComponent(filteredProps as T, ref);
       }
       (filteredProps as { ref?: unknown }).ref = ref;
+      if (typeof Component === "string" && "as" in filteredProps) {
+        const { as: Tag, ...rest } = filteredProps as { as: string };
+        return <Tag {...rest} />;
+      }
       return <Component {...(filteredProps as any)} />;
     };
     return yakForwardRef(yak, mergedAttrsFn);
--- a/packages/next-yak/runtime/styled.tsx
+++ b/packages/next-yak/runtime/styled.tsx
@@ -214,6 +214,7 @@ const yakStyled = <
         return parentYakComponent(filteredProps as T, ref);
       }
       (filteredProps as { ref?: unknown }).ref = ref;
+      // allow to overwrite the html tag with `as`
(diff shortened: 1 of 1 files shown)


jantimon:
Closed for now - see #184

PR #195, opened 2024-10-30
Use minimal FunctionComponent

### LINKED ISSUE #194: styled() does not accept "generic" function components
## Description

The `styled` function from next-yak fails to accept a generic component when the component uses an intersection type with React.FunctionComponent. The error occurs because the type definition creates an incompatibility between the component's type signature and React's FunctionComponent type requirements.

```tsx
interface MyGenericProps<T> {
  input: T
}

type MyGenericComponent<T> =  React.FunctionComponent<MyGenericProps<T>> & {
  <TOther>(props: MyGenericProps<TOther>): React.ReactNode
}

type MyDefaultType = {
  [x: string]: any
}

const GenericComponent:  MyGenericComponent<MyDefaultType> = () => {
  return null;
}

const ConcreteStyledComponent = GenericComponent<"test">

// should allow a "incorrect" function component
const MyComponent = styled(ConcreteStyledComponent)``;

// should allow input as "test"
<MyComponent input="test" />
```
[playground](https://www.typescriptlang.org/play/?jsx=4#code/JYWwDg9gTgLgBAbwM4wJ4BsCmATAvnAMyghDgCIA7TADxgFpUBDAazIG4AoUSWOAJUyMAxvCIlyUQSPYcuFGJigFhmOAFlUAcUxUowIQAViYJAB4AKgD5EHOHGAUwAVxgAuOOY65ZaMKo3auvoAwiSQVPIW1gC8dgLCMAB0AGJOFCLAEBSh4Fk6MKYBOor6RhAmUdYAZDZ2FgDyMAAWipYAFGDGSO5FQYZdDc2tAJTu8SKJ4zAAchDYmF4+qH7qqAAimMpO6DDmy6qxCLZwANrU7ih6FADmALrujBSoixxCWShwgSVCOeH57qsvnofmE8pENBstjs9n4YnA2sM4NFrEc7JIYE4oBQ4BRtuhON5Xu94KF0ujMABlNBYbC-MHwWJAkKgiIFMgKFBkSyyAD0PLgSCaEG22DgjHQ6AgAHcxeQHG8oJJpIQ0hksnA3rlWUSKB8NHTWUiBdScG1SUJyVSMDgDflhgADe2cDh8gVCkViiXS+yOFxipDkDkwMgcQqoW3yH3OGDRdmYTlwHmWIA)

## Current behavior
There is a type error when declaring the styled component: `styled(ConcreteStyledComponent)`.

## Expected behavior
The `styled` function accepts the concrete generic component.

## Root Cause
This issues root cause is that the generic component type definition has two different generics that get inferred/assigned at two different time points. `MyGenericComponent` sets `MyDefaultType` as default type and the generic flows to the `FunctionComponent` and is set. Once the `GenericComponent` is created with a new type `"test"` this flows to the props of the function and is therefore incompatible with the `FunctionComponent` definition of react:
```ts
    interface FunctionComponent<P = {}> {
        (props: P, context?: any): ReactNode;
        propTypes?: WeakValidationMap<P> | undefined;
        contextTypes?: ValidationMap<any> | undefined;
        defaultProps?: Partial<P> | undefined;
        displayName?: string | undefined;
    }
```

The function will be overridden to `(props: MyGenericProps<"test">): React.ReactNode` while the `propTypes` and `defaultProps` still be `WeakValidationMap<MyDefaultType>` respectively `Partial<MyDefaultType>`.

## Root Cause 2
But there is another issue. This code works with `styled-components` because they typed the function better. `next-yak` expects a valid `FunctionComponent` even tough a "normal" function would be enough, as we don't use any special properties of the `FunctionComponent` type. 

## Proposition
I would propose to fix the issue in `next-yak` and making the required type area smaller in order to be compatible with this generic component.

Mad-Kat:
Fixes #194 

Instead of using reacts `FunctionComponent` with all the additional properties, `next-yak` works with the minimal function declaration.


jantimon (approved):
nice - why do we need the context?

PR #210, opened 2024-11-13
[docs] Fix SWC transform in vercel function

Mad-Kat:


jantimon (approved):

Mad-Kat:
I think I know why the `benchmarks` check never finishes. The benchmark action only runs on `next-yak`, `yak-swc` or `benchmark` changes but the check is always required. So if we have a PR that doesn't update any path of them, the check never succeeds. Either we should make the check only required for these paths (if that is possible) or make it optional for all PR's

Issue #208, opened 2024-11-12
Css template literal variable scope issue


christinaheidt:
Currently it is not allowed to use anything other than a constant and it's therefore generating an error as followed:
![image](https://github.com/user-attachments/assets/1419962e-4282-4909-9dff-eb6c78b32b42)
![image](https://github.com/user-attachments/assets/195a8a35-65f8-46a0-99a3-2824456ca39b)

Mad-Kat:
That means number 2 works, but number 1 doesn't? And it's a feature request not a bug report? 😄 

What is the expectation in this case if the variable exists in the top scope? Should we use that or the "local" property?

jantimon:
If I understand correctly, you're running into an issue with nested CSS template literals accessing variables from an outer arrow function scope. 

This is happening because the `$pageHeaderHeight` value is being accessed directly inside a nested `css` template literal, which isn't supported by next-yak currently. 

Here's a minimal example that demonstrates the same issue:

```tsx
const Button = styled.div<{ $active: boolean }>`
  ${({ $active }) => css`
    margin-top: ${$active}px;  // Error: The value for variable "$active" could not be found in the top scope
  `}
`
```

The only workaround right now is an additional arrow function to properly scope these variables. 

Here's how you can fix it:

```tsx
const Button = styled.div<{ $active: boolean }>`
  ${({ $active }) => css`
    margin-top: ${() => $active}px;
  `}
`
```

We should improve next-yak's handling of this case. The current error message "The value for variable could not be found in the top scope" is misleading since the variable exists but can't be accessed and "top scope" is hard to understand

We could either improve the error message or modify the compiler to handle this case.

Usually we enforce an function for css-variable values so that when someone reads the code it is clear which part will become a css-variable and which part can be inlined as constant

PR #212, opened 2024-11-17
Improve dynamic value error message

Mad-Kat:
Closes: #208

Error message before:
```
  x The value for variable "$pageHeaderHeight" could not be found in the top scope
    ,-[input.js:17:1]
 16 |     ` : css`
 17 |       margin-top: ${$pageHeaderHeight}px;
    :                     ^^^^^^^^^^^^^^^^^
 18 |     `};
    `----
```

Error message afterwards:
```
  x The shorthand access to the variable "$pageHeaderHeight" is not allowed in a nested expression.
  | To be able to use the property turn it into a CSS variable by wrapping it in a function:
  | 
  | ${() => $pageHeaderHeight};
  | 
    ,-[input.js:17:1]
 16 |     ` : css`
 17 |       margin-top: ${$pageHeaderHeight}px;
    :                     ^^^^^^^^^^^^^^^^^
 18 |     `};
    `----
```


jantimon (approved):
it could have a false positive if there is a typo in a constant name - but I guess that’s handled by typescript anyway

PR #220, opened 2024-11-25
add own class name hashing

jantimon:
This PR is a first draft for minified CSS class names in production by generating deterministic but unique class names directly in the SWC plugin

The change will:

- reduce the class name size in css file (upto 20% gzip file size savings)
- reduce the ssr html size (because the class names are also used there)
- reduce the js bundle size (because we can skip the mappings and the class names are also there)

The generated code from the SWC plugin changes from

```tsx
import styleYak from "./input.yak.module.css"; 
styled.button(styleYak.Button);

to the following:

```tsx
import "./input.yak.module.css";
styled.button("Button_m7uBBu");
```

The PR will only work for Next.js versions with a `postcss-modules-local-by-default` version 4.1 or newer 

Still todo:
- [x] Test if single line comments `//` work for postcss ignore markers
- [ ] Update css-loader imports from cross-file selectors and mixins to use the new class naming scheme


packages/next-yak/loaders/css-loader.ts
```diff
@@ -43,6 +43,11 @@ function extractCss(code: string): string {
     const codeUntilEnd = codeParts[i].split("*/")[0];
     result += codeUntilEnd;
   }
+  // postcss can't handle // comments
+  result = result.replace(
+    /\/\/ cssmodules-pure-ignore/g,
+    "/* cssmodules-pure-ignore */",
```
Mad-Kat, line 49: Can't we do this directly in the SWC plugin?

packages/yak-swc/yak_swc/src/naming_convention.rs
```diff
@@ -153,7 +188,16 @@ mod tests {
...
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU2");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU3");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU4");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU5");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU6");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU7");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU8");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbU9");
+    assert_eq!(convention.get_css_variable_name("foo"), "yoPBkbUA");
+    assert_eq!(convention.get_css_variable_name(""), "yoPBkbUB");
```
Mad-Kat, line 200: Would be nice to see what happens if the values will reach >16

packages/yak-swc/yak_swc/tests/fixture/keyframe-animations-function/output.tsx
```diff
@@ -21,7 +24,7 @@ const fadeIn = /*YAK Extracted CSS:
     opacity: 1;
   }
 }
-*/ /*#__PURE__*/ keyframes(__styleYak.fadeIn);
+*/ /*#__PURE__*/ keyframes("fadeIn");
```
Mad-Kat, line 27: Does this work when the keyframe is local to the module?

Mad-Kat (changes requested):

packages/yak-swc/yak_swc/src/naming_convention.rs
```diff
@@ -118,6 +133,26 @@ fn escape_css_identifier(input: &str) -> String {
   result
 }
 
+/// Convert a number to a CSS-safe string
+fn minify_number(num: u32) -> String {
```
Mad-Kat, line 137: add specific tests

packages/yak-swc/yak_swc/tests/fixture/keyframe-animations-function/output.tsx
Mad-Kat: todo: add `:global()` to the keyframe name

packages/next-yak/loaders/css-loader.ts
jantimon: Unfortunately not as `*/` will end the outer comment

packages/yak-swc/yak_swc/tests/fixture/keyframe-animations-function/output.tsx
jantimon: fixed

jantimon:
This PR requires

next 15.3.0-canary.30 (for CI pipelines)
next 15.3.0 (for stable release)

packages/yak-swc/yak_swc/src/naming_convention.rs
jantimon: done

packages/yak-swc/yak_swc/src/naming_convention.rs
```diff
@@ -43,6 +45,27 @@ impl NamingConvention {
...
   }
 
+  /// Returns the base name of the file
+  /// This is used during development to help with debugging
+  pub fn get_file_name_base(&mut self) -> String {
+    if let Some(base) = &self.file_name_base {
+      base.clone()
+    } else {
+      // remove file extension and path \\ and /
+      let base = self
```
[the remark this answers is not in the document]
chklauser: If, and only if, we are talking about a path in the developer's file system. If `self.file_name` is a "logical" path in a source file import statement, then I would stick with `String`.

packages/next-yak/loaders/css-loader.ts
```diff
@@ -62,6 +62,10 @@ function extractCss(code: string | Buffer<ArrayBufferLike>): string {
     const codeUntilEnd = codeParts[i].split("*/")[0];
     result += codeUntilEnd;
   }
+  if (result) {
+    result = "/* cssmodules-pure-no-check */\n" + result;
```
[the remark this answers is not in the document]
jantimon: unfortunately we can't place comments inside comments as `*/` would terminate the generated css comment by the swc plugin

packages/yak-swc/yak_swc/src/naming_convention.rs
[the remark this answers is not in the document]
jantimon: I didn't choose `path` because I believe it is a normalized path (always unix) and therefore might not work well with Path on windows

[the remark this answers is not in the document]
Mad-Kat: They [fixed it for the absolut path](https://github.com/rust-lang/rust/issues/92750), but as we're only interested in [the stem](https://doc.rust-lang.org/std/path/struct.Path.html#method.file_stem) I don't think this should be a problem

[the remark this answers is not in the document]
jantimon: we need also to strip the path e.g. `/foo/bar.tsx` -> `bar` and this might be broken for windows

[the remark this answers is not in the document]
jantimon: I changed it to `use std::path::Path;` 👍

PR #225, opened 2024-12-03
allow to define the prefix for generated names

jantimon:
I've added support for configuring the prefix used by next-yak for generated CSS classes and variables  
This can help in identifying yak-generated classes in tests or to avoid collisions with other libraries

The default prefix remains `y` in production and empty in dev mode, so this is fully backwards compatible

If needed you can now configure a custom prefix:

```js
// next.config.js
export default withYak({
  prefix: 'my-app-'
});
```

The patch also adjusts the dev mode variable prefix from 'yak_' to 'var_' which makes a bit more sense semantically

### DIFF
--- a/.changeset/silver-hairs-clean.md
+++ b/.changeset/silver-hairs-clean.md
@@ -1,0 +1,5 @@
+---
+"yak-swc": minor
+---
+
+allow to define the prefix for generated names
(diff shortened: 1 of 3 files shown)


.changeset/silver-hairs-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+"yak-swc": minor
+---
+
+allow to define the prefix for generated names
```
Mad-Kat, line 5:
```suggestion
Add possibility for a custom class name prefix
```

Mad-Kat (approved):

.changeset/silver-hairs-clean.md
jantimon: for now it's only the custom css variable name - in future (#255) it will also be the class name

PR #227, opened 2024-12-03
allow unquoted urls

### LINKED ISSUE #226: CSS Parser fails to handle unquoted URLs in background properties
@laem just encountered an issue with the CSS parser when using unquoted URLs in background properties:

```css
background: url(https://example.com/image.png);
```

in this case the parser goes wild and in the end the minifier throws an "Unclosed block" error

However, it works fine when the URL is wrapped in quotes:

```css
background: url("https://example.com/image.png");
```

according to the CSS spec both forms are valid

I think the parser is interpreting the colon in the URL as a CSS property-value separator when it's not wrapped in quotes

See https://github.com/jantimon/next-yak/issues/113#issuecomment-2514621888

jantimon:
previously, the CSS parser would incorrectly interpret special characters (like colons) inside unquoted CSS function values as syntax separators. This caused the parser to break when encountering valid CSS like:

```css
background: url(https://example.com/image.png);
```

this fix adds parenthesis depth tracking to properly handle all CSS function values (url, calc, rgb, etc.) regardless of whether they are quoted

Fixes #226


Mad-Kat:
Would it make sense to add a fixture case for the bug that is addressed?

Mad-Kat (dismissed):

Mad-Kat (approved):

PR #230, opened 2024-12-04
Improve the error message for dynamic css functions when returning invalid values

### LINKED ISSUE #229: Undefined dynamic prop gives misleading error message
```
//Component
return (						
<Group key={group} >...</Group> // missing $groupColor={groupColor}
)


const Group = styled.li`
	border-left: 4px solid ${(p) => p.$groupColor};
	padding-left: 0.4rem;
`


```  
image

It took me a while to understand that the problem was an undefined prop. I was looking for some place in my code where CSS was written without being enclosed by the `css` template literal function. 

Solution : change the message for something like "<GroupColor styled component has a missing prop"

jantimon:
When a styled component's dynamic prop function returns an invalid value (like `undefined` from a missing prop), the error message now includes the actual function that caused the issue.

Before:
```
Error: Dynamic CSS functions must return a string or number but returned undefined
```

After:
```
Error: Dynamic CSS functions must return a string or number but returned undefined

Dynamic CSS function: (p) => p.$groupColor
```

Fixes #229


Mad-Kat:
A test case with the actual error message would be nice for the next time we encounter this. Wdyt?

Mad-Kat (approved):

PR #231, opened 2024-12-06
Split up YakImportVisitor from YakImports collections

Tyderion:
Add new struct that splits up visitor from the collectiosn which are used to store the used imports in yak.
This makes it easier to add/transform custom imports (for e.g. #154)

The given structs&collections are also private with only the needed functions exposed to use them


packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -137,6 +137,13 @@ where
     self.current_css_state.is_some()
   }
 
+  fn yak_imports(&self) -> &YakImports {
+    self
+      .yak_library_imports
+      .as_ref()
+      .expect("If this is reached the imports should've been parsed already.")
```
jantimon, line 144:
```suggestion
      .expect("Internal error: yak_library_imports is None - this should be impossible as imports are parsed in the initial program visit before any other processing")
```

jantimon (approved):

PR #233, opened 2024-12-06
Use AsRef<str> instead of String for parameters/arguments

Tyderion:
This enables functions to be called with Strings or string slices


packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -361,9 +361,9 @@ pub struct TransformKeyframes {
 }
 
 impl TransformKeyframes {
-  pub fn new(animation_name: String) -> TransformKeyframes {
+  pub fn new(animation_name: impl AsRef<str>) -> TransformKeyframes {
```
jantimon, line 364: why do we need to change a transformer?

Tyderion: we don't need to, i just thought to make it consistent (i.e. all functions accept string slices as well as strings as args)

jantimon (changes requested):
in that case I would drop that transformer change (for now)

jantimon (approved):

Issue #240, opened 2024-12-17
Error when using the CSS prop with an object 


Henrixounez:
When using the CSS Prop as an object, as shown in the [docs here](https://yak.js.org/features#css-prop), it fails to compile and give a TypeError

## ✅   Working Example with the CSS Utility Function:

```tsx
  <p
    css={css`
      color: green;
    `}
  >
    CSS Prop works if this is green
  </p>
```
![image](https://github.com/user-attachments/assets/a0564521-08a7-4706-a7ac-5cae51d69558)

## ❌  Not working using as an object:

```tsx
  <p
    css={{
      color: "green"
    }}
  >
    CSS Prop works if this is green
  </p>
```
![image](https://github.com/user-attachments/assets/f5e3dfa6-c029-4366-a6d2-0b3952807011)

```
 ⨯ TypeError: {(intermediate value)} is not a function
    at Home (app/page.tsx:144:10)
  142 |           css={{
  143 |             color: "green"
> 144 |           }}
      |          ^
  145 |         >
  146 |           CSS Prop works if this is green
  147 |         </p> {
  digest: '2695420628'
}
 GET / 500 in 73ms
```

jantimon:
oh you are right - objects are not supported

@Mad-Kat can remember why we added the css prop docs like this?

currently the css prop only works like this:

`<div css={css`color:green`}>`

Mad-Kat:
Thanks a lot for posting this :) We're always happy for input.

> @Mad-Kat can remember why we added the css prop docs like this?

I don't remember exactly. I assume it worked with our babel plugin, but doesn't work currently with the SWC plugin. I really need to overhaul the docs. 🙈 

> I was also wondering if it was possible to have dynamic values inside the css props instead of having to create a dedicated > styled variable with props, when needed for quick inline styling on tags like whats possible with styled-components / emotion ?

We implemented the css prop in a very limited scope. It only works on native elements and without any interpolation. To support the interpolation we would have to extract a styled component for you in the correct scope the same way as `styled-components` does it. I'm not a huge fan of it, as it makes the code way more error prone and it's a complex task.

I assume your problem is mainly that your codebase is full with the css prop and it's very tedious to migrate? Or what is the use case here? Is it only for the light-dark theming or do you need to have other dynamic properties?

Mad-Kat:
And btw: Thank you for the very clear issue

Henrixounez:
[...]

PR #242, opened 2024-12-18
Enable conditional styling for the css prop

Mad-Kat:
This was noticed in #240 

With our SWC plugin it should be possible to allow dynamic css properties to a certain extend. 

So with this PR this works:

```tsx
const show = Math.random() > 0.5;
<div
  css={css`
    ${() =>
      show &&
      css`
        padding: 20px;
      `}
  `}
/>
```

and also css variables:

```tsx
const show = Math.random() > 0.5;
<div
  css={css`
    padding: ${() => show && '20px'};
  `}
/>
```

Whereas dynamic css props with their own properties still don't work:

```tsx
<div
  css={css<{$show: boolean}>` // this won't work
    ${({$show}) =>
      show &&
      css`
        padding: 20px;
      `}
  `}
/>
```

and also css props that use mixins that define their own props aren't allowed either:

```tsx
const myMixin = css<{$show: boolean}>`
  ${({$show}) =>
    show &&
    css`
      padding: 20px;
    `}
`;
<div
  css={css`
    ${myMixin} // this won't work
  `}
/>
```


.changeset/red-rats-double.md
```diff
@@ -0,0 +1,6 @@
+---
+"next-yak": patch
+"yak-swc": patch
+---
+
+Allow dynamic props in css prop
```
jantimon, line 6:
```suggestion
Enable conditional styling for the css prop
```

packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -191,17 +191,16 @@ impl YakTransform for TransformCssMixin {
   ) -> YakTransformResult {
     let has_dynamic_content = !runtime_expressions.is_empty() || !runtime_css_variables.is_empty();
 
-    if (self.is_exported || self.is_within_jsx_attribute) && has_dynamic_content {
+    if self.is_exported && has_dynamic_content && !self.is_within_jsx_attribute {
       // For now dynamic mixins are not supported cross file
```
jantimon, line 195: which case would that be?
is it easy to add a fixture test for this error?

jantimon (dismissed):

.changeset/red-rats-double.md
Mad-Kat: We already got a fixture that covers this:
https://github.com/jantimon/next-yak/tree/main/packages/yak-swc/yak_swc/tests/fixture/cross-file-mixin-dynamic-export

Do you think this covers the case enough or should I write another case?

jantimon (approved):

PR #243, opened 2024-12-19
Enable css props on styled components

Mad-Kat:
This was noticed in #240 

This change alters the public API of styled components e.g. `styled.div` or `styled.p` to accept an optional CSS property the same way as native elements do with our JSX definition.

This works now without a type error:

```tsx
const Text = styled.p`
  font-size: 20px;
  font-weight: bold;
  font-family: "Inter";
`;

<Text
      css={css`
        color: red;
      `}
    >
      test
    </Text>
```

### DIFF
--- a/.changeset/angry-dingos-remain.md
+++ b/.changeset/angry-dingos-remain.md
@@ -3,4 +3,4 @@
 "yak-swc": patch
 ---
 
-Fix types to enable css prop on styled components
+Enable css prop support for styled components, not just native HTML elements by fixing a bug in the types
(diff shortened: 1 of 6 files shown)


.changeset/angry-dingos-remain.md
```diff
@@ -0,0 +1,6 @@
+---
+"next-yak": patch
+"yak-swc": patch
+---
+
+Fix types to enable css prop on styled components
```
jantimon, line 6:
```suggestion
Enable css prop support for styled components, not just native HTML elements by fixing a bug in the types
```

jantimon (approved):

jantimon (approved):

PR #249, opened 2024-12-29
Use pnpm catalog

Mad-Kat:
With version 9.5.0, pnpm added the ability to specify the versions of all packages centrally. I've run the codemod and changed the minimum version to 9.7.0, as they added the ability to [manage package manger versions](https://pnpm.io/npmrc#manage-package-manager-versions) that we already specify in our root `package.json`.


jantimon (changes requested):
We need a way to prevent `npm publish` otherwise `dependencies: { "yak-swc": "catalog:" }` would be published to npm and break on install

Also we might consider splitting the catalog into our core dependencies (which will change the behaviour once we publish to npm) and a dev dependency catalog (which includes docs, vitest, ...)

jantimon (approved):

PR #253, opened 2024-12-30
Fix css prop in exported component

### LINKED ISSUE #252: CSS prop doesn't work when component that uses it is exported
When we have a CSS prop in a component that itself is exported e.g.

```tsx
import { css } from "next-yak";

export const Yak = () => (
  <div
    css={css`
      display: flex;
      gap: 6px;
    `}
  >
    Yak
  </div>
);
```

the compilation doesn't add the class name to it:

```jsx
import { css, __yak_mergeCssProp } from "next-yak/internal";
import __styleYak from "./input.yak.module.css!=!./input?./input.yak.module.css";
export const YakLogo = ()=><div {.../*YAK EXPORTED MIXIN:YakLogo
.YakLogo {
  display: flex;
  gap: 6px;
}
*/ /*#__PURE__*/ css()({})}>
    Yak
  </div>;
```

and this means that the styling isn't applied.

Mad-Kat:
Closes #252 

Handling a css prop should take precedence over exported mixins.

### DIFF
--- a/packages/yak-swc/yak_swc/src/yak_transforms.rs
+++ b/packages/yak-swc/yak_swc/src/yak_transforms.rs
@@ -218,17 +218,6 @@ impl YakTransform for TransformCssMixin {
       );
     }
     let css_prefix = match (self.is_exported, self.is_within_jsx_attribute) {
-      (true, _) => Some(format!(
-        "YAK EXPORTED MIXIN:{}",
-        self
-          .export_name
-          .as_ref()
-          .unwrap()
-          .parts
-          .iter()
-          .map(|atom| encode_percent(atom.as_str()))
-          .join(":")
-      )),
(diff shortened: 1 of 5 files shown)


jantimon (dismissed):

packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -218,17 +218,6 @@ impl YakTransform for TransformCssMixin {
...
-        self
-          .export_name
-          .as_ref()
-          .unwrap()
-          .parts
-          .iter()
-          .map(|atom| encode_percent(atom.as_str()))
-          .join(":")
-      )),
       (_, true) => {
```
jantimon, line 221: an `if` would be easier to read than a match

PR #254, opened 2025-01-02
Update docs

Mad-Kat:
Updated docs

- Gave the landing page a make over
- Styled almost all with `next-yak`
- Overhauled the migration guide
- Streamlined the rest

Preview: https://next-yak-docs-9njrd6pud-jan-nicklas-projects.vercel.app/


packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
 ---
 title: Migration from styled-components
-description: We tried to keep the changes you need to know if you're already familiar with styled-components to a minimum. 
+description: We tried to keep the changes you need to know if you're already familiar with styled-components to a minimum.
```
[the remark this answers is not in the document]
jantimon: maybe we could also mention that this was one of the core goals of yak so we can switch from styled-components without losing all the know-how and dev tooling

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
-```json
-{
-	"compilerOptions": {
-      "jsxImportSource": "next-yak"
-	}
-}
+<Callout>
+**Important note on semicolons**:
+You need to add a semicolon after the mixin to separate it from the rest of the styles.
+Otherwise it will be treated as a part of the styles and our parser has no way to differentiate between the two.
```
jantimon, line 113:
```suggestion
Otherwise the parser can not differentiate between a selector, a constant or a mixin.
```

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
+</Callout>
+
+### Attrs
```
jantimon, line 116: Move after keyframes?

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
 ```
-</Tab>
-</Tabs>
 
-### References to other components
+## Move some code to yak files
 
-In order to create a selector that includes another component, you can just reference it in the tagged
-template literal. 
+Other usages of styled-components need a bit more work. Yak files are special files that are evaluated during build time and can be used to store static values that are extracted during build time. You can create a `.yak.ts` file everywhere in your project and it will be picked up by next-yak.
```
jantimon, line 244:
```suggestion
Migrating `styled-components` with complex logic might need a bit more work. Yak files are special files that are evaluated during build time and can be used to store static values that are extracted during build time. You can create a `.yak.ts` or `.yak.tsx` file everywhere in your project and it will be picked up by next-yak.

<Callout>
**Important note on `.yak.tsx` files**:
As yak files are evaluated on every code change, they will slow down your dev experience and your production build times. They should only be used as a last resort.
</Callout>
```

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
-way as you would with styled-components.
+### Shared values
 
-<Tabs groupId="frameworks" items={["styled-components", "next-yak"]}>
-<Tab value="styled-components">
-```jsx
-import styled, { css } from 'styled-components';
+Another use case for yak files is to store shared values that are used across your application.
+Next-yak tries to optimize each file individually and doesn't share values between files.
+So if you have a value that is used in multiple files, you should move it to a yak file so that it can be optimized across all files.
```
jantimon, line 279: We should explain that everything but calculations and string concatenations work WITHOUT yak files and should be preferred to keep the compilation fast

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
 
-## Migration
+## Change some more things
 
-### Change import
+The last part of the migration guide is about more complex cases where you need to change a bit more than just the import.
 
-Change the import from `styled-components` to `next-yak` and change the import of `styled` from a default import 
-to a named import.
+### CSS prop
```
jantimon, line 306:
```suggestion
### CSS prop (optional)
```

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
+        background: papayawhip;
+        color: red;
+      `}
+    />
+  );
 };
 ```
 
-You can then use these values in your styled components.
+Unfortunately, you can't use dynamic values in the css prop. If you need to use dynamic values, you should create a styled component with the styles from the css prop.
```
jantimon, line 343: Instead show an example how dynamic runtime props work here

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
 
-```tsx
-import { styled } from 'next-yak';
-import { colors, breakpoints } from './constants.yak';
+### Dynamic component styles
```
jantimon, line 345: Maybe we should move this up (before yak files) under a Performance section

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
+```tsx title="transformed.tsx"
+// pseudo code
 
-```tsx
-import styled, { css, keyframes } from 'styled-components'; // [!code --]
-import { styled, css, keyframes } from 'next-yak'; // [!code ++]
+const Button = (props) => props.$primary
+  ? <button className="next-yak-1">Click me</button>
+  : <button className="next-yak-2">Click me</button>
+```
```
jantimon, line 402:
```suggestion
const Button = (props) => props.$primary
  ? <button className="next-yak-1">Click me</button>
  : <button className="next-yak-2">Click me</button>
  
  
const Button = (props) => 
  <button className={props.$primary ? "next-yak-1" : "next-yak-2"}>
     Click me
  </button>
```
```

File: packages/docs/content/docs/migration-from-styled-components.mdx [hunk 10]
```diff
@@ -1,514 +1,582 @@
...
 import { styled } from 'next-yak';
 
 const Button = styled.button<{ $primary: boolean }>`
-	background: #BF4F74;
-	color: ${props => props.$primary ? "white" : "#BF4F74"};
+  background: #BF4F74;
+  color: ${props => props.$primary ? "white" : "#BF4F74"};
+
+  /* Alternative syntax */
+  ${props => props.$primary ? "color: white;" : "color: #BF4F74;"}
```
jantimon, line 83: here ``css`...` `` is missing - which is also important when migrating - should we also mention this somewhere?

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
+```tsx title="util.ts"
+export const generateAllColors = (primaryColor: string) => { // [!code --]
+  return { // [!code --]
+    primary: primaryColor, // [!code --]
+    secondary: "#F7B801", // [!code --]
+  }; // [!code --]
+}; // [!code --]
+```
+
+```tsx title="util.yak.ts"
```
jantimon, line 505: I don't think this needs to be a `.yak` file

packages/docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -1,514 +1,582 @@
...
+export const allColors = { // [!code ++]
+  "red": { // [!code ++]
```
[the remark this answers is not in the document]
jantimon: I believe it's a constant in this case - so that should be fine

jantimon (approved):

PR #264, opened 2025-01-19
Docs: Improve landing page

Mad-Kat:
- Adds an animated code example (instead of the video) but keeps the video for smaller screens
- Adds a performance section with the measured benefits
- Adds basic metadata to all pages
- Improves image rendering

Preview: https://next-yak-docs-ac8qjgmim-jan-nicklas-projects.vercel.app/


jantimon:
I would prefer the animations to happen sequentially so that the user can see the changes step by step

Also in the video it slowly first removes e.g. styled-components to give the user time to look at the right place before next-yak comes in - can we find some way to highlight the place which will change?

jantimon (approved):

jantimon:
Updated docs:
https://next-yak-docs-mtj7uv1zx-jan-nicklas-projects.vercel.app/

PR #274, opened 2025-02-12
Improve css-loader cache to avoid failing build with nextjs and webpack

devbrains-com:
After migrating more and more styled-components usages to next-yak, we started to experience failed nextjs builds (next build). The build ended without any error message and with a success status code, which is quite strange. After some time analyzing the issue, we were able to relate the issue to the parsed files cache within the css-loader.

Initially, the cache stored the promise resolving and parsing a module and its dependencies. A parallel request resolving the same module would, therefore, await the same promise and not load the module again. This likely led to a deadlock, or it has just exceeded the maximum number of concurrent loader executions.

Removing the cache solved the issue but slowed the build down quite a bit. With this PR, we change the cache to only cache the final result with the compromise of having the same module resolved more than once if requested simultaneously. Our performance test showed the exact same build time but without any failing builds.

### DIFF
--- a/packages/docs/content/docs/how-does-it-work.mdx
+++ b/packages/docs/content/docs/how-does-it-work.mdx
@@ -23,7 +23,11 @@ the runtime part. The runtime part is responsible for merging styles and props.
 `next-yak` uses an SWC plugin and an additional webpack loader to transform your code. The plugin is responsible for transforming the
 usages of the tagged template literals (like `styled` and `css`), while the loader reads the results from the plugin, resolves cross module dependencies and writes the finished css in a css-modules file.
 
+<<<<<<< Updated upstream
 ### The SWC plugin
+=======
+### The first loader [(yak-swc)](https://github.com/jantimon/next-yak/blob/main/packages/next-yak/loaders/ts-loader.ts)
+>>>>>>> Stashed changes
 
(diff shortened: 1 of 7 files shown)


jantimon (approved):

Mad-Kat:
@jantimon looks like the secrets aren't injected into the actions when a PR builds from a branch on a fork. Can/should we enable this? Otherwise we don't know if it would break the deployment of the example and docs.

packages/docs/content/docs/how-does-it-work.mdx
```diff
@@ -23,7 +23,11 @@ the runtime part. The runtime part is responsible for merging styles and props.
 `next-yak` uses an SWC plugin and an additional webpack loader to transform your code. The plugin is responsible for transforming the
 usages of the tagged template literals (like `styled` and `css`), while the loader reads the results from the plugin, resolves cross module dependencies and writes the finished css in a css-modules file.
 
+<<<<<<< Updated upstream
```
Mad-Kat, line 26: This looks like a merge error

Mad-Kat (changes requested):

Mad-Kat:
@devbrains-com Thanks for hunting down the bug. However I think a little description about the bug and how this PR resolves it might be handy for future references. Could you add that?

Mad-Kat (approved):

PR #282, opened 2025-03-04
drop codspeed because of excessive permission request

jantimon:
codspeed

I asked the codspeed but didn't get any answer


KrzysztofKarol (approved):

jantimon:
maybe @adriencaccia knows why codspeed tries to gain read/write access to our Administration and Actions

adriencaccia:
> maybe @adriencaccia knows why codspeed tries to gain read/write access to our Administration and Actions

Hey @jantimon, it should be explained in the developers notes: 
image

> CodSpeed now registers self-hosted runners directly on repositories instead of organizations for wall time runs.
> 
> Learn more in the [documentation](https://docs.codspeed.io/features/walltime/).

This is the API call that we use: https://docs.github.com/en/rest/actions/self-hosted-runners?apiVersion=2022-11-28#create-configuration-for-a-just-in-time-runner-for-a-repository

adriencaccia:
> I asked the codspeed but didn't get any answer

Sorry about that; we must have missed your message. Where did you send it?

jantimon:
@adriencaccia thank you so much!
we will definitely try to setup codspeed up again and give feedback

PR #296, opened 2025-03-25
Minor file visitor refactoring

jeremystucki:
The diff looks enormous, but it really is mostly reduced nesting. If you look at it [without whitespaces changes](https://github.com/DigitecGalaxus/next-yak/pull/296/files?diff=split&w=1) it looks a lot better.

Feel free to close without merging if this causes too much noise.


.changeset/smart-timers-attack.md
```diff
@@ -0,0 +1,2 @@
+---
```
jantimon, line 1: this is production change - could you add a small changeset?

e.g. Improve Transpilation Performance

packages/yak-swc/yak_swc/src/yak_file_visitor.rs
```diff
@@ -59,85 +59,85 @@ impl VisitMut for YakFileVisitor {
     // This is necessary as the mixin is also imported at runtime and a string would be
     // interpreted as a class name
     if let Expr::TaggedTpl(n) = expr {
-      if let Some(name) = self
+      if let Some("css") = self
```
jantimon, line 62: this looks strange - can we really hardcode this to css?

jeremystucki: In other visitors we also match atoms against string literals, so I assumed that would be fine.

.changeset/smart-timers-attack.md
jeremystucki: I don't think this has a performance impact, it just makes the code more readable. Would you still add a changeset?

packages/yak-swc/yak_swc/src/yak_file_visitor.rs
Mad-Kat: In general it would be nice to always check it via the `yak_library_imports` as someone could import the function an alias them. E.g.

```jsx
import {css as YakCSS} from "next-yak"

const Mixin = YakCSS`
  color: red;
`;
```

.changeset/smart-timers-attack.md
Mad-Kat: Yes please :)

packages/yak-swc/yak_swc/src/yak_file_visitor.rs
jeremystucki: Was this the case with the `name == atom!("css")` comparison this replaces?

Mad-Kat: Unfortunately, I don't think so. Maybe something we should assess generally? I will create an issue for it

jeremystucki: Thank you, I'll resolve this conversation then since this PR does not change that behavior.

jeremystucki:
FYI I force-pushed to use my DG email address.

jantimon (approved):

Mad-Kat (approved):
Thank you.

PR #302, opened 2025-03-31
Make devMode configurable

### LINKED ISSUE #301: Make `devMode` configurable
There is a "hidden" config setting `devMode` that controls some aspects of code generation. Today it is always set to
```js
process.env.NODE_ENV !== "production";
```

That's a reasonable default, but users might want to control this setting explicitly. We propose to expose the setting in `YakConfigOptions` and fall back to `NODE_ENV` if it is unset.

chklauser:
Resolves #301


chklauser:
I experimented a bit with fumadocs [typescript](https://fumadocs.vercel.app/docs/ui/typescript) support (to automatically generate docs based on typescript types). I wasn't exactly blown away by the result. 
* Default value is assumed to be code (formatted as `<code>`). Can't have prose these :(
* Can't hide the `experiments` type (I would want to document it separately)

Before I invest a bunch of time into documenting the config options by hand, I would want to be sure that this is something we want to do (document it twice, once in tsdoc and once in mdx).

jantimon:
Maybe we need to explain what `devMode` does - i quickly asked AI and it summarized it like this:

### When `devMode` is `true`:

1. **More readable class names and variables**:
   - The naming convention generates longer, more descriptive class names
   - Includes the original component name in generated CSS classes
   - For example, `Button` might become `Button-01` instead of just a hash

2. **No prefix by default**:
   - In the `NamingConvention::new` method, when `dev_mode` is true, it uses an empty string as the default prefix
   - This makes class names more readable in development

3. **CSS variables include the original property name**:
   - When generating CSS variable names, it includes the property name for easier debugging
   - For example: `Button__color_oPBkbU` instead of just `yoPBkbU`

4. **Counter format is more verbose**:
   - When generating unique names, it uses a more verbose format like `-01`, `-02` instead of just `1`, `2`

### When `devMode` is `false` (production mode):

1. **Optimized for production**:
   - Generates shorter, minified class names
   - Uses "y" as the default prefix
   - Omits descriptive parts to reduce the size of the generated CSS

2. **Minimal CSS variable names**:
   - CSS variables are shorter and lack the descriptive components
   - This reduces the overall bundle size

3. **More compact counter format**:
   - Uses simple numeric suffixes like `1`, `2` instead of `-01`, `-02`

So maybe we should rename it to `minifyNames` (or maybe a even better name?) this will become even more important once we can merge https://github.com/DigitecGalaxus/next-yak/pull/220

jantimon:
lets wait for this pr until #220 is merged

jantimon:
@chklauser #220 is merged - do you have time to change devMode into minify? otherwise I can easily pick it up

chklauser:
I like the `minify` idea. Have added a commit that switches the config setting around.

jantimon (approved):

PR #307, opened 2025-04-08
Use css to merge properties instead of internal function

Mad-Kat:
This PR removes the internal `mergeCssProp` utility and instead uses the css function directly to merge props, updating both the compiler output and runtime behavior.

- Drop `__yak_mergeCssProp` from SWC output and adjust transformation logic in Rust to call `css(...)({...})`.
- Remove the merged runtime helper and inline merging via `combineProps`/`mergeClassNames`
- Update tests, example usage, and change logs to reflect the new CSS prop strategy.

This doesn't change the public API for `css`

**Note:** This changes behavior slightly. As the logic of merging class names moved to the css function, when using something like this

```jsx
<div className="myClassName" css={css`
  ${() => css``};
  ${() => css``};
`}>
```

the resulting HTML looks like this:

```html
<div class="yakCss myClassName yakCss1 myClassName yakCss2" />
```

The original class string is multiplied by the number of css interpolations inside the css prop.

Should we deduplicate it (e.g. mark the outermost css function somehow and only it should add the class names passed from the props)?


packages/next-yak/runtime/cssLiteral.tsx
```diff
@@ -97,22 +99,17 @@ export function css<TProps>(...args: Array<any>): ComponentStyles<TProps> {
...
   }
 
-  // Non Dynamic CSS
-  if (dynamicCssFunctions.length === 0) {
-    const className = classNames.join(" ");
-    return () => ({ className, style });
-  }
-
-  return (props: unknown) => {
+  // @ts-ignore
```
jantimon, line 102: better use ts-expect-error and explain why

packages/next-yak/runtime/styled.tsx
```diff
@@ -25,7 +26,7 @@ const noTheme: YakTheme = {};
  */
 type HtmlTags = keyof React.JSX.IntrinsicElements;
 
-/**
```
jantimon, line 28: ??

packages/next-yak/runtime/cssLiteral.tsx
```diff
@@ -55,15 +55,17 @@ export function css<TProps>(
...
+): (props: TProps) => TProps {
   const classNames: string[] = [];
   const dynamicCssFunctions: PropsToClassNameFn[] = [];
   const style: Record<string, string> = {};
   for (const arg of args as Array<string | CSSFunction | CSSStyles<any>>) {
     // A CSS-module class name which got auto generated during build from static css
     // e.g. css`color: red;`
     // compiled -> css("yak31e4")
-    if (typeof arg === "string") {
+    if (typeof arg === "string" && arg.length !== 0) {
```
jantimon, line 68: in which case do we need `css("")`?

Mad-Kat:
Superseded by https://github.com/DigitecGalaxus/next-yak/pull/375

PR #308, opened 2025-04-09
Support `atoms` in CSS prop

Mad-Kat:
This supersedes #291 

> Completing this PR will add the possibility to use atoms directly in the css prop. E.g.
```jsx
<div
  css={atoms('flex gap-4', isSomething ? 'basis-8/12' : 'w-full')}
/>
```

Based on #307 the change to add the atoms functionality is trivial, as the transpilation step can stay the same and only the runtime API of the `atoms` function has to be changed.

### DIFF
--- a/.changeset/ninety-tips-attack.md
+++ b/.changeset/ninety-tips-attack.md
@@ -1,0 +1,6 @@
+---
+"next-yak": patch
+"yak-swc": patch
+---
+
+Added support for using the atoms function directly in the css prop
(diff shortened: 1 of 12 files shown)


Mad-Kat:
Sorry for the delay. I will find time this or next week to discuss it once more with @jantimon. I just want to ensure that the API is good enough

Mad-Kat:
@therealgilles Sorry for creating another PR 😄 

But I think we find a very promising solution based on the idea mentioned above. 

If you're interested #375 is to be followed. Once I think the idea is in a good shape, I will close the other PR's that try to tackle the same problems.

Sorry for the delay.

Mad-Kat:
Superseded by https://github.com/DigitecGalaxus/next-yak/pull/375

PR #314, opened 2025-04-11
fix cross-file selectors by using embedded class names

### LINKED ISSUE #213: Import of selector fails when it hasn't any base styles
When importing components as selectors, we always attempt to generate an import for the component's base class, which causes issues with components that only have dynamic classes based on props.

## Currently

When using an imported component as a selector:

```jsx
import { Label } from "./other";

const OverwriteLabel = styled.div`
  ${Label}{
    color: pink;
  }
`
```

we generate:

```jsx
const OverwriteLabel = /*YAK Extracted CSS:
.OverwriteLabel {
  --yak-css-import: url("./other:Label",selector) {
    color: pink;
  }
}
*/ /*#__PURE__*/ styled.div(__styleYak.OverwriteLabel);
```

which results in this CSS:

```css
.OverwriteLabel {
  :global(.other_Label__l1FMk) {
    color: pink;
  }
}
```

## Problem

The generated CSS assumes there's a base class named Label in the imported component's CSS. However, this assumption fails when the imported component only has dynamic classes based on props:

```jsx
export const Label = styled.label<{ $variant: "primary" | "secondary" }>`
  ${({ $variant }) => $variant === "primary" ? css`
    color: red;
  ` : css`
    color: blue;
  `}
`;
```

which results in this CSS:

```css
.Label__ {
  color: red;
}
.Label__1 {
  color: blue;
}
```

In this case, there is no base .Label class to target, causing the selector to fail.

## Proposed Solutions

1. Generate Base Class Always
    Always generate a base class for components, regardless of whether they have static styles
    **Pros**: Maintains backwards compatibility
    **Cons**: Additional CSS overhead

2. Throw Descriptive Error
  Detect when a component lacks a base class and throw an error with an explanation of the issue and suggestions for solutions like "add base styles to the component" or "split into separate components"
  **Pros**: Clearer developer experience
  **Cons**: Breaking change for the migration from `styled-components` to `next-yak`

jantimon:
I've fixed the cross-file selector issue that was occurring because we're now generating our own class names rather than using Next.js's `getCssModuleLocalIdent`.

Instead of trying to replicate the class name generation logic across both the Rust and JavaScript sides (which would be error-prone), I've taken the same approach we use for mixins. For exported styled components, we now embed the component name and its generated class name directly in the CSS comment, using the format `/*YAK EXPORTED STYLED:ComponentName:ClassName*/`.

The CSS loader now parses these comments to extract the styled component information, giving us perfect consistency between the class names used in the component's file and when it's referenced from another file.

This approach only generates the metadata for exported components since those are the only ones that might be referenced from other files. It's also fully compatible with our existing minification and prefix features since the class name is pre-generated on the Rust side.

I've tested this with various exported component references, including default exports and named exports, and all cross-file references now seem to work as expected

fixes #213

### DIFF
--- a/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
+++ b/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
@@ -689,6 +689,11 @@ type ParsedExport =
   | { type: "star-export"; from: string[] };
 
 type ResolvedExport =
-  | { type: "styled-component"; from: string; name: string; value: string | undefined }
+  | {
+      type: "styled-component";
+      from: string;
+      name: string;
+      value: string | undefined;
+    }
   | { type: "mixin"; value: string | number }
   | { type: "constant"; value: string | number };


packages/cross-file-tests/__tests__/fixtures/nestedMixin/output/index.yak.module.css
```diff
@@ -1,23 +1,14 @@
 /* cssmodules-pure-no-check * /
 :global(.index_Button_oPtJMe) {
   color: black;
-:global(.icon_Icon__vdEsc) {
+--yak-css-import: url("./icon:Icon",selector) {
```
Mad-Kat, line 4: This can't be good. It looks like resolving the chain `mixin -> mixin & selector -> source` doesn't work correctly

Mad-Kat (changes requested):
Looks almost perfect

Mad-Kat:
I've pushed a fix. Please have a look if it makes sense

jantimon:
> Would this also fix https://github.com/DigitecGalaxus/next-yak/issues/213?

yes - it was a low hanging fruit

Mad-Kat (approved):

packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
```diff
@@ -657,6 +688,6 @@ type ParsedExport =
   | { type: "star-export"; from: string[] };
 
 type ResolvedExport =
-  | { type: "styled-component"; from: string; name: string }
```
jantimon, line 660: I have to look into this - actually this should not happen..

the case where value is `undefined` is when we parse the file with babel - however all exported styled-component information are gathered from the extracted comment

maybe this happens because of reexports?

jantimon: i checked and the only way might be to change the type of `ParsedFile` to always have a value

Mad-Kat (approved):

PR #318, opened 2025-04-14
fix SWC plugin compatibility with Turbopack

jantimon:
This PR addresses an issue where Next.js Turbopack crashes when using SWC plugins configured with absolute paths

## Problem

When using `require.resolve()` or other methods that return absolute paths in the `swcPlugins` configuration, Turbopack fails with a module resolution error. This happens because Turbopack tries to resolve these absolute paths as relative paths, causing the build to crash immediately

Error example:
```
Module not found: Can't resolve './path/to/node_modules/yak-swc/target/wasm32-wasip1/release/yak_swc.wasm'
```

Instead of using `resolve("yak-swc")` which returns an absolute path (and works in weback) we can directly use `"yak-swc"`

### DIFF
--- a/.changeset/early-dingos-lie.md
+++ b/.changeset/early-dingos-lie.md
@@ -1,0 +1,6 @@
+---
+"next-yak": minor
+"yak-swc": minor
+---
+
+use a turbopack compatible swc plugin path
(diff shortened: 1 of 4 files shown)


Mad-Kat:
Our test setup has some issues with the change. Could you have a look?

jantimon:
> Our test setup has some issues with the change. Could you have a look?

the problem was that `yak-swc` was not a dependency of cross-file-tests and therefore could not be resolved

I hope that this is a mono repository issue as `yak-swc` is a dependency of `next-yak`

Mad-Kat (approved):

PR #327, opened 2025-04-19
Hoist @property rules to root level

### LINKED ISSUE #325: Can CSS Houdini styles be used?
Before
```
  @property --angle {
    syntax: '<angle>';
    inherits: true;
    initial-value: 0turn;
  }

  @keyframes spin {
    to {
      --angle: 1turn;
    }
  }

  animation: spin 3s linear infinite;
```
It works in @emotion/styled
Now it's not working, need help~

jantimon:
This PR addresses issue #325 where CSS Houdini's `@property` rules were not working correctly when nested inside selectors. According to CSS specs, `@property` rules must appear at the top level of a stylesheet to function properly.

The fix **modifies the CSS stringification** to identify `@property` rules and hoist them to the root level, maintaining compatibility with other CSS processors like Emotion/styled-components that automatically handle this hoisting.

**Testing**: Added test case to verify `@property` hoisting behavior.

Fixes #325

### DIFF
--- a/packages/yak-swc/css_in_js_parser/src/to_css.rs
+++ b/packages/yak-swc/css_in_js_parser/src/to_css.rs
@@ -1,6 +1,29 @@
 //! Converts a list of CSS declarations to a CSS string
 use crate::{CssScope, Declaration, ScopeType};
 
+/// Converts a list of CSS declarations into properly formatted CSS code
+///
+/// This function is handling indentation, scope nesting, and maintaining proper selector hierarchy
+/// It does that by combining declarations that share the same scope
+///
+/// # Example
+///
+/// ```
+/// let declarations = vec![
+///   Declaration {
+///     property: "color".to_string(),
+///     value: "blue".to_string(),
+///     scope: vec![
+///       CssScope {
+///         name: ".foo".to_string(),
+///         scope_type: ScopeType::Selector
+///       }
+///     ]
+///   }
+/// ];
+/// let css_string = to_css(&declarations);
+/// // Results in: ".foo { color: blue; }"
+/// ```
 pub fn to_css(declarations: &[Declaration]) -> String {
   let mut hoisted_css = String::new();
   let mut regular_css = String::new();
--- a/packages/yak-swc/css_in_js_parser/src/to_css.rs
+++ b/packages/yak-swc/css_in_js_parser/src/to_css.rs
@@ -1,18 +1,46 @@
 //! Converts a list of CSS declarations to a CSS string
-use crate::{CssScope, Declaration};
+use crate::{CssScope, Declaration, ScopeType};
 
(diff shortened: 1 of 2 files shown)


packages/yak-swc/css_in_js_parser/src/to_css.rs
```diff
@@ -38,12 +66,33 @@ pub fn to_css(declarations: &[Declaration]) -> String {
...
+  // Close all regular scopes with proper indentation
   for i in (0..previous_scopes.len()).rev() {
-    css.push_str(&format!("\n{}}}", "  ".repeat(i)));
+    regular_css.push_str(&format!("\n{}}}", "  ".repeat(i)));
   }
 
-  css
+  // Combine hoisted CSS and regular CSS
+  if !hoisted_css.is_empty() && !regular_css.is_empty() {
+    format!("{}{}", hoisted_css, regular_css)
```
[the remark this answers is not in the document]
jantimon: as you can see in the test that's already the case

packages/yak-swc/css_in_js_parser/src/to_css.rs
```diff
@@ -38,12 +66,33 @@ pub fn to_css(declarations: &[Declaration]) -> String {
...
+  } else {
+    hoisted_css + &regular_css
+  }
+}
+
+/// If according to the css spec the inner scope of declaration must not be nested
+/// (e.g. @property) return that scope
+fn get_non_nestable_declarations(declaration: &Declaration) -> Option<CssScope> {
+  for scope in &declaration.scope {
+    if scope.scope_type == ScopeType::AtRule && (scope.name.starts_with("@property")) {
```
Mad-Kat, line 91: Should we also do this for keyframes?

packages/yak-swc/css_in_js_parser/src/to_css.rs
```diff
@@ -1,18 +1,46 @@
 //! Converts a list of CSS declarations to a CSS string
-use crate::{CssScope, Declaration};
+use crate::{CssScope, Declaration, ScopeType};
 
 pub fn to_css(declarations: &[Declaration]) -> String {
```
Mad-Kat, line 4: A small documentation what this function does would be nice

Mad-Kat (approved):
Just some ideas for improvement, but it works and can be merged at your discretion

packages/yak-swc/css_in_js_parser/src/to_css.rs
jantimon: yes but it would require some tweaking

`@property --foo {` has 1 scope

but

`@keyframes { from {` has 2 scopes

that's why I didn't implement it right away but used variable/function names which allow such an extension

Mad-Kat (approved):

PR #329, opened 2025-04-23
Create eslint-plugin

Mad-Kat:
This PR will create an ESLint plugin to help with the migration from `styled-components` to `next-yak`.

The following rules are created:

- `css-nesting-operator` Enforces css selectors in next-yak to correctly use the nesting selector (&)
- `enforce-semicolon` Enforces that expression in styled/css literals from next-yak use semicolons
- `style-conditions` Enforces that arrow functions only return runtime values or css literals in styled/css literals from next-yak


packages/eslint-plugin/README.md
```diff
@@ -0,0 +1,57 @@
+# eslint-plugin-yak
+
+ESLint plugin for [next-yak](https://yak.js.org/) - a powerful CSS-in-JS library for React applications.
+
+Helps with the migration from `styled-components` to `next-yak`.
```
jantimon, line 5: Maybe also - Hints you about performance optimizations

Mad-Kat: Do you think this is good enough?

Mad-Kat:
Does the release work with a new package that doesn't need to be published in the same interval as `next-yak` and `yak-swc`?

jantimon (dismissed):

packages/eslint-plugin-yak/docs/rules/css-nesting-operator.md
```diff
@@ -0,0 +1,73 @@
+# Enforces css selectors in next-yak to correctly use the nesting selector (&) (`yak/css-nesting-operator`)
+
+🔧💡 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix) and manually fixable by [editor suggestions](https://eslint.org/docs/latest/use/core-concepts#rule-suggestions).
+
+<!-- end auto-generated rule header -->
+
+Enforces css selectors in next-yak to correctly use the [nesting selector (&)](https://developer.mozilla.org/en-US/docs/Web/CSS/Nesting_selector). Unlike in `styled-components`, the nesting selector is required to be used in order to correctly scope styles.
```
jantimon, line 7: Maybe we can also write the reason for our decision

e.g. (if I remember correclty) here we said that we don't auto inject the `&` so that it's closer to vanilla css and works well with the css nesting spec

packages/eslint-plugin-yak/docs/rules/enforce-semicolon.md
```diff
@@ -0,0 +1,67 @@
+# Enforces that expression in styled/css literals from next-yak use semicolons (`yak/enforce-semicolon`)
+
+🔧 This rule is automatically fixable by the [`--fix` CLI option](https://eslint.org/docs/latest/user-guide/command-line-interface#--fix).
+
+<!-- end auto-generated rule header -->
+
+Enforces semicolons after a mixin, to make distinguishing between mixins and nested selectors easier.
```
jantimon, line 7: Maybe we can also write the reason for our decision

e.g. Background: runtime css-in-js libraries merge all strings at runtime - however as next-yak doesn't have a runtime it must combine mixins at build time. without a semicolon it is hard for next-yak to understand the difference between a variable which holds a slector e.g. ``css`${something} { color: blue }` `` and an mixin e.g. ``css`${skeleton}```

packages/eslint-plugin-yak/docs/rules/style-conditions.md
```diff
@@ -0,0 +1,79 @@
+# Enforces that arrow functions only return runtime values or css literals in styled/css literals from next-yak (`yak/style-conditions`)
+
+💭 This rule requires [type information](https://typescript-eslint.io/linting/typed-linting).
+
+<!-- end auto-generated rule header -->
+
+Warns if runtime performance could be improved by using css literals.
```
jantimon, line 7: Maybe we can add some background

e.g.: There are multiple ways to define dynamic styles. Most dynamic styles are a variations are only turned on/off and can be precompiled into their own css class like in the following example:

```js
styled.button<{$primary?: boolean}>`
  color: blue;
  ${({$primary}) => $primary && css`
        color: red;
  }`
`;
```

color:blue is part of the base css class and ``css`color:red` `` can be statically extracted into another css class

The second approach allows to be 100% dynamic at runtime but has a small overhead and should therefore be only used for cases which are not static. For example:

js```
styled.div<{$x: number}>`
  left: ${x}px
`;
```

In the second example there the value for `x` is statically extracted into a css variable e.g. `--x`.
The consequence is that the variable has to be set in the html like this: `<div style="--x:43px">...</div>`

The `yak/style-conditions` rule tries to report unnecessary variable usages to keep your html smaller

jantimon (approved):

jantimon (approved):

PR #342, opened 2025-05-09
Comparison

jantimon:


Mad-Kat:
I think we should add a header row that categorizes the libraries in: "build-time", "run-time". And we should order the columns after popularity or popularity and category. What do you think?

Mad-Kat (approved):

PR #343, opened 2025-05-09
Playground with wasm-compiled swc

chklauser:
# Approach
Integrating the yak_swc plugin into the WASM build of swc is awkward because there is no straightforward way to have the wasm plugin infrastructure work when swc itself is already executing in web assembly.

We instead use code that is essentially a copy of the existing swc `build_transform_sync` macro, which lets us register custom before/after passes. Instead of integrating yak as a plugin, we configure it manually as a "before" pass.

# Concequences
We need to build `yak_swc` without the swc plugin infrastructure. In the swc crates there is a "plugin mode". When it is enabled, "real" swc functionality gets replaced with plugin stubs that are expected to be provided from outside the sandbox. That won't work for us as we want our code to use the real swc implementation that we are compiling into the same, shared wasm image. 
The current solution is to feature-gate the plugin integration functionality behind an enabled-by-default feature of `yak_swc`. The playground build disables that feature.

A "production" build takes quite a while (2-4 minutes), but we definitely want a production build for a web deployment. An unoptimized WASM image is ~80MB.

```
[INFO]: 🌀  Compiling to Wasm...
   Compiling swc_core v22.5.2
   Compiling yak_swc v0.1.0 (/home/chris/devel/next-yak/packages/yak-swc/yak_swc)
   Compiling playground-wasm v0.1.0 (/home/chris/devel/next-yak/packages/playground-wasm/wasm-build)
    Finished `release` profile [optimized] target(s) in 47.88s
[INFO]: ⬇️  Installing wasm-bindgen...
[INFO]: Optimizing wasm binaries with `wasm-opt`...
[INFO]: Optional fields missing from Cargo.toml: 'description', 'repository', and 'license'. These are not necessary, but recommended
[INFO]: ✨   Done in 1m 38s
[INFO]: 📦   Your wasm pkg is ready to publish at wasm.
```

In terms of size, I could reduce the image to 10MB. GZipped, that's roughly 2.5MB.
```
───┬─────────────────────────┬──────┬──────────┬─────────────────────
 # │          name           │ type │   size   │      modified
───┼─────────────────────────┼──────┼──────────┼─────────────────────
 0 │ wasm/index.d.ts         │ file │ 58.8 KiB │ 2025-05-09 14:48:26
 1 │ wasm/index.js           │ file │ 19.7 KiB │ 2025-05-09 14:48:26
 2 │ wasm/index_bg.wasm      │ file │ 10.5 MiB │ 2025-05-09 14:49:16
 3 │ wasm/index_bg.wasm.d.ts │ file │    686 B │ 2025-05-09 14:48:26
 4 │ wasm/package.json       │ file │    235 B │ 2025-05-09 14:49:16
───┴─────────────────────────┴──────┴──────────┴─────────────────────
```

# TODO 

- [x] Integrate into website
- [ ] Show diagnostics
- [ ] ...?

# Ugly PoC
 image


jantimon:
Finally - a working playground!

The compilation speed is crazy!

And also cross file resolution works 🤯

3 small things:

- scrolling doesn't work in the output
- output doesn't wrap
- the file name of the tab is not used for the compilation but "theFile"

chklauser:
I have
* forwarded the `filename` set in the options object passed to the transform fn
* added the logic to treat `*.yak.*` files as, well, yak files (as long as the correct `filename` gets passed from the JS-side of the playground, it should just work

jantimon (approved):

PR #344, opened 2025-05-10
Set regexp engine to forgiving for safari

Mad-Kat:
From the [shiki docs](https://shiki.style/guide/regex-engines#use-with-unsupported-languages):

> Unlike the Oniguruma engine, the JavaScript engine is strict by default. It will throw an error if it encounters an invalid Oniguruma pattern or a pattern that it cannot convert. If you want best-effort results for unsupported grammars, you can enable the forgiving option to suppress any conversion errors.

This enables `forgiving` mode for our highlighter to work in safari.

Preview: https://next-yak-docs-77rr3m2km-jan-nicklas-projects.vercel.app/


jantimon (approved):
super cool

PR #345, opened 2025-05-11
more docs

jantimon:
i tried to add some of juhos feedback


packages/docs/components/landingPage.tsx
```diff
@@ -70,46 +81,72 @@ export const LandingPage = () => {
...
       <Subtitle>Performance</Subtitle>
       <p
         style={{
           marginBottom: "1rem",
         }}
       >
-        Validated across many thousands of real-world users:
+        Next-Yak is way faster than most other CSS-in-JS libraries. Learn more
+        about it's{" "}
+        <NextLink href="/docs/how-it-works">
```
Mad-Kat, line 94:
```suggestion
        <NextLink href="/docs/how-does-it-work">
```

Mad-Kat: And should we have an underline for these links to be able to differentiate between links and bold statements?

packages/docs/components/landingPage.tsx
```diff
@@ -260,3 +297,41 @@ const EnumTitle = styled.div`
 const Svg = styled.svg`
   height: 0.7lh;
 `;
+
+const VersionLink = styled.a`
+  display: inline-block;
+  font-size: 1.2rem;
+  padding: 0.2rem 0.7rem;
+  border-radius: 9999px;
+  background: rgb(37 99 235);
```
Mad-Kat, line 306: The color looks strange in light mode next to the cta-button:
image

packages/docs/components/landingPage.tsx
```diff
@@ -260,3 +297,41 @@ const EnumTitle = styled.div`
...
+    background-position-x: 100%;
+  }
+`;
+
+const Strong = styled.strong`
+  ${theme.dark} {
+    background: linear-gradient(45deg, #fc00ff, #00dbde, #fc00ff) -100%/ 200%;
+    background-clip: text;
+  }
+  background: linear-gradient(45deg, #5a005b, #005355, #5a005b) -100%/ 200%;
```
Mad-Kat, line 331: In my opinion, the shimmer could pop more in the light mode: 
image

Mad-Kat (changes requested):

packages/docs/components/landingPage.tsx
jantimon: I also wasn't happy with the color - do you have a better color in mind?

packages/docs/components/landingPage.tsx
jantimon: yes looks boring - do you want to play around a little bit?

Mad-Kat: I've changed it. Is this better? What do you think?

PR #357, opened 2025-05-24
Handle ParsedExports in nested records

### LINKED ISSUE #356: Cross-file constant resolution fails for nested objects beyond one level
The webpack loader fails to resolve cross-file constants when accessing nested object properties beyond one level deep.

## Reproduction

Create a file with nested constants:

```ts
// someFile.ts
export const tokens = { 
  spacing: { 
    s1: "1rem" 
  } 
};
```

Try to access the nested value in another file:

```tsx
import { tokens } from "./someFile";

const Button = styled.button`
  padding: ${tokens.spacing.s1}
`;
```

## Expected Behaviour

The padding should be set to "1rem".

Instead it fails with:

```text
Error: Error resolving cross-file selectors: Error resolving from module ...example/app/someFile.ts: Error unpacking Record/Array - could not extract s1 from tokens.spacing

Error: Error resolving cross-file selectors: Error resolving from module ...example/app/someFile.ts: Error unpacking Record/Array - could not extract `s1` from `tokens.spacing`
Extracted values: {
  "tokens": {
    "type": "record",
    "value": {
      "spacing": {
        "type": "record",
        "value": {
          "s1": {
            "type": "constant",
            "value": "1rem"
          }
        }
      }
    }
  }
}
```

---
Thank you @aberba for [mentioning this on X](https://x.com/aberba/status/1925804300795568299)

Mad-Kat:
This fixes #356 

We didn't handle the case where the exported record object already contained `ParsedExports`. I've added the handling, added a small visual test case in our example app and refactored the record unwrapping logic.

### DIFF
--- a/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
+++ b/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
@@ -514,6 +514,24 @@ function parseObjectExpression(
   return result;
 }
 
+function handleNestedRecordValue(
+  value: any,
+  specifier: string[],
+): ResolvedExport {
+  if (typeof value === "string" || typeof value === "number") {
+    return { type: "constant" as const, value };
+  } else if (value && typeof value === "object") {
+    if ("__yak" in value) {
+      return { type: "mixin", value: value["__yak"] };
+    } else if ("value" in value) {
+      return { type: "constant", value: value["value"] };
+    }
+  }
+  throw new Error(
+    `Error unpacking Record/Array - could not extract final value from \`${specifier.join(".")}\``,
+  );
+}
+
 /**
  * Follows a specifier recursively until it finds its constant value
  * for example here it follows "colors.primary"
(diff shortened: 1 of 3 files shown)


packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts
```diff
@@ -632,28 +651,23 @@ async function resolveModuleSpecifierRecursively(
...
           );
         }
-        depth++;
-        // mixins in .yak files are wrapped inside an object with a __yak key
-        if (depth === specifier.length && "__yak" in current) {
-          return { type: "mixin", value: current["__yak"] };
-        } else if (depth === specifier.length && "value" in current) {
-          return { type: "constant", value: current["value"] };
-        } else {
-          current = current[specifier[depth]];
```
jantimon, line 642: this line is actually the line which resolves nested values

can we fix it?

jantimon (approved):

PR #385, opened 2025-07-14
Add support for default exported components

rvetere:


packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -889,14 +940,54 @@ where
...
+          // Normal case: add the full comment
+          self.comments.add_leading(
+            result_span.lo,
+            Comment {
+              kind: swc_core::common::comments::CommentKind::Block,
+              span: DUMMY_SP,
+              text: format!("{}\n{}\n", comment_prefix, css_code.trim()).into(),
+            },
+          );
+        }
```
Mad-Kat, line 990: Instead of replacing the comments after the creation, I think this logic should be handled by `yak_transforms.rs`

packages/yak-swc/yak_swc/src/utils/yak_constants.rs
```diff
@@ -0,0 +1,11 @@
+/// Prefix for exported styled components
+/// Format: "YAK EXPORTED STYLED:{export_name}:{class_name}"
+pub const YAK_EXPORTED_STYLED_PREFIX: &str = "YAK EXPORTED STYLED:";
+
+/// Prefix for exported CSS mixins  
+/// Format: "YAK EXPORTED MIXIN:{export_name}"
+pub const YAK_EXPORTED_MIXIN_PREFIX: &str = "YAK EXPORTED MIXIN:";
+
+/// Prefix for extracted CSS without export metadata
+/// Used for non-exported components and as fallback
```
Mad-Kat, line 10: This is used for all types of build time CSS (not just as fallback or on non-exported components)

Mad-Kat:
@rvetere An additional case we missed:

```jsx
export default Title;

var Title = styled.h1`
  color: blue;
  font-size: 24px;
`;
```

What if the default export is before the variable assignment?

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -630,10 +696,19 @@ where
...
+        // Check if this variable will be default-exported
+        let will_be_default_exported = self.variables.is_default_exported(&id.to_id());
+        
         self.current_variable_name = Some(ScopedVariableReference::new(
           id.to_id(),
           vec![id.sym.clone()],
         ));
+        
+        if will_be_default_exported {
+          self.current_exported = true;
```
Mad-Kat, line 709:
```suggestion
          self.current_exported = self.variables.is_default_exported(&id.to_id());
```

And the if + the temporary variable aren't used anymore

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -800,6 +875,7 @@ where
 
     let is_top_level = !self.is_inside_css_expression();
     let current_variable_id = self.get_current_component_id();
+    let is_default_export = self.current_exported && self.variables.is_default_exported(&current_variable_id.id);
```
Mad-Kat, line 878: Is there a world where `self.current_exported` is false but `self.variables.is_default_exported` is true?

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -889,14 +967,39 @@ where
...
+        // Check if this is a default-exported styled component or mixin with separated comments
+        if let Some(css_only_comment) = transform_result.css.css_only_comment.clone() {
+          // Store the full comment for later use at export default statement
+          let full_comment = format!("{}\n{}\n", comment_prefix, css_code.trim());
+          self.default_export_styled_components.insert(
+            current_variable_id.id.0.to_string(),
+            full_comment
+          );
+          
+          // Add only the CSS comment for the variable declaration
```
Mad-Kat, line 981: Can't we add here the full comment as well (to reduce special case handling) just without the "default" name?

packages/yak-swc/yak_swc/src/variable_visitor.rs
```diff
@@ -22,6 +23,10 @@ pub enum ImportSourceType {
 pub struct VariableVisitor {
   variables: FxHashMap<Id, Box<Expr>>,
   imports: FxHashMap<Id, ImportKind>,
+  /// Variable name that will be default-exported (only one per module)
+  default_exported_variable: Option<Id>,
+  /// Span of the default export statement (for adding comments)
+  default_export_span: Option<Span>,
```
Mad-Kat, line 29: We could use a Tuple here to capture the Span and the Id

packages/yak-swc/yak_swc/src/variable_visitor.rs
```diff
@@ -138,6 +145,21 @@ impl VariableVisitor {
     }
     None
   }
+
+  /// Returns true if the given variable name will be default-exported
+  pub fn is_default_exported(&self, name: &Id) -> bool {
+    self.default_exported_variable.as_ref().map_or(false, |var| var == name)
```
Mad-Kat, line 151:
```suggestion
    self.default_exported_variable.as_ref() == Some(Id);
```

packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -18,6 +19,8 @@ use crate::naming_convention::{NamingConvention, TranspilationMode};
 #[derive(Debug)]
 pub struct YakCss {
   pub comment_prefix: Option<String>,
+  /// For default exports, this contains only the CSS comment without export prefix
```
Mad-Kat, line 22: It would reduce complexity if we could use the same comment.

Mad-Kat (changes requested):

Issue #373, opened 2025-06-24
Non exported styled component with no rules gets no class


PaulBlanche:
Repro in the playground : https://yak.js.org/playground?q=H4sIAMyYWmgCA1VRvWoDMQzW7KcQB4VkSO1Ct7tmyd6pULqdm5himlwO1y0HJe3WJkP2kCV5Fj1Z5PuLz2ChT5a%2BT5IVaKAzG4N8BND%2Fk%2FVLA7S9lUCnJBUClkB7%2BMFvoL%2BRCGnZzBTeuGkNGC7sV%2BsiAu16f6WrCdAB75Uqq7QN55vW6cqZwK7eeoD44eYPCdgkCsWsMfOdUjcd8YAcUUb89UzTR1P5yYt%2Bz2SDr%2B%2FPTpelcX1JJnmkBmTyOuw4FRvexzwPW6HjbF14bQvj4JfT8xA1Vbl2ntvdt5Sj8WBvXUUv1HTigx20xbJxbq0cvurE5pXVNXzyXVwAi7wfh78BAAA%3D

If you explore the markup, you see that the `div` corresponding to `Container` does not get a `class`.

This is an issue because even if `Container` is not exported, it might be used as a selector in the same file.

Issue #370, opened 2025-06-16
unable to resolve default component


PaulBlanche:
Given the following files : 
```ts
//Title.ts
const Title = styled.h1`
  color: red;
`;

export default Title
```
```ts
//Card.ts
import Title from "./Title.ts"

const Card = styled.div`
  ${Title} {
    color: blue;
  }
`
```

The following css is generated :
```
.title-fgy45 {
  color: red;
}

.title-rt5bY {
  undefined {
    color: blue;
  }
}
```

`next-yak` do not "understand" that the `Title` component is exported (probably because of the default export indirection), and the `/*YAK EXPORTED STYLED:...*/` comment is missing (the comment with the extracted CSS is there though).

This means that the loader can't resolve `Title` in `Card.ts` as a styled component. Instead there is only a styled-component with no className (since it was not generated by `next-yak`), so we end up with an `undefined` in the css.

[...]

Issue #369, opened 2025-06-04
unable to resolve styled component in exported object


PaulBlanche:
Here we merge exported styled component with the export object : 
If we have this input : 
```ts
const S = {
  regular: styled``
}
```

We get the following comment from `yak-swc` : 
```ts
  /*YAK EXPORTED STYLED:S.regular:yUvTHDB9Q*/
```

This is parsed as a `styled-component` with a name `S.regular`. When merging with the export object we have a bug : https://github.com/DigitecGalaxus/next-yak/blob/main/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts#L257-L263

The export object for `S` will contains a record with a property `regular`, whereas the parsed style component is a flat object with dotted keys `S.regular`.

This means the className of the parsed style component will not be affected to the property inside the `S` record, and using this component as a selector will fail (the css will contain `undefined` instead of the className).

The same logic applied to mixins (https://github.com/DigitecGalaxus/next-yak/blob/main/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts#L265-L306) should also be applied to styled components

jantimon:
Hello @PaulBlanche 

could you please explain once again the problem?

I tried to reproduce your bug but it seems to work just fine in the playground:

[...]

Issue #368, opened 2025-06-04
`yak-swc` not handling static computed key


PaulBlanche:
Given the following code : 
```ts
type AvatarSizes = 'extra-small' | 'small' 

const styles = {
  'extra-small': css`
    height: 30px;
    width: 30px;
  `,
  small: css`
    height: 36px;
    width: 36px;
  `,
} satisfies Record<AvatarSizes, any>

type VariantProps = {
  $size?: AvatarSizes
}

const dynamicAvatar = css<VariantProps>`
  ${({ $size }) => {
    if ($size === 'extra-small') {
      return styles['extra-small']
    }
    if ($size === 'small') {
      return styles.small
    }
  }}
`

export const s = styled.div<VariantProps>`
  ${dynamicAvatar}
`
```

`yak-swc` (version 5.6.1) throws no error on `styles['extra-small']` but is not able to infer that it references a mixin on `styles` (so no class for the mixin is outputed), whereas `styles.small` is correctly handled (a class is outputed).

I think `yak-swc` interpret `styles['xxx']` as a computed key and skips it. It should see the expression `'extra-small'` is a static literal and handle it like an identifier. I also got the same problem when using array indexing like `styles.foo[1]`

The issue can be sidesteped by using valid identifier as keys in `styles` so the issue is not blocking for me, but i thought i'd report it anyway.

jantimon:
Thanks for the feedback 

You are right - `styles['extra-small']` is a computed expression in SWC but you are right it's static so I believe we could add support for that case

Here is a reproduction of that bug in the playground:

[...]

PR #389, opened 2025-07-21
Make the cross-file resolve logic usable with different bundlers

PaulBlanche:
Next-yak is a great project, but in its current form it is limited to a specific bundler: webpack. If we don't use nextjs or webpack it is possible to use `yak-swc` directly in our bundler of choice, but all the logic in the "cross-file" loader must be reimplemented, because it is tightly connected to the webpack api.

This PR propose a refactor of the cross-file logic to abstract away any logic coming from webpack : 
- module resolving
- module loading to get the code content
- module parsing to get the AST
- module transform to get the `yak-swc` transformed code
- module evaluation for `.yak` files
- caching of the operations above.

As an nice side effect, the logic become much more testable, and i added some unit tests for basic cases and some edge cases.

### `parseModule.ts`

Given a path, `parseModule` returns a `ParsedModule` containing : 
- the exports of the module (grouped by named export and all exports)
- the type `"regular"` or `"yak"` (for `.yak` files)
- the transformed code via `yak-swc` if the module import `next-yak`
- all exported mixins and styled component (parsed from the comments left by `yak-swc`)

This operation can be cached. Instead of caching the promise we cache the value it resolves to, because from what i understood, it could cause deadlock (but since there is no tests for this case, i have now way to know if the problem persists).

### `resolveCrossFileConstant.ts`

Given a path and some css, `resolveCrossFileConstant` returns the css with the `--yak-css-import(...)` replaced with the resolved value.

Here again we can cache the work done, with a "twist" : the result of `resolveCrossFileConstant` depends on the content of multiple files, so i added logic to track the file dependencies of a "resolve" operation. This dependency list can be used by the cache to invalidate entry : 
- with esbuild a call to `fs.stat` for each dependencies of a value before accessing it in cache let us now if the value is fresh or stale
- with rollup (and rolldown ?) we can get the list of files that changed between two builds to invalidate the cache.
- for webpack i don't know if such an optimisation is possible, so i left the operation uncached

### Interface

The cross-file logic relies on multiple functions : 
- `resolve`, to resolve a module given its specifier and the importer
- `evaluateYakModule`, to evaluate a `.yak` file. This method is optionnal, and not providing it disable `.yak` file handling.
- `extractExports` to get the exports of a module. Given that you use `babel`, but could use `swc` and `rollup` as its own AST, i decided to extract this operation to be compatible with any AST.
- `getTransformed` to get the code transform with `yak-swc`. Here again the logic differs between webpack  (wich relies on a `yak` loader beeing present) and esbuild/rollup where the transform would happen within the plugin

### Optimisations

Appart from the new caching "level" on the resolve stage, i also added an "optional" bail-out when the module does not import `yak`. This bailout is not implemented here (the implemented `extractExports` always returns `importYak: true`) because i'm not sure it would make a great difference with webpack (but in my experience it does with esbuild).

:warning: I tried adding some config to `tsup` to output the cross-file logic and expose it as `next-yak/resolver` or something like that, but no matter how i do it, either the import path in the compiled js do not match, or the types in the `.d.ts` are not there :/
So this is currently missing in my PR


Mad-Kat:
Hey @PaulBlanche 

Thank you so much for this PR. You're working on the part that is tested the least, complex and hard to understand. Nevertheless you tackled it and I like the direction you're taking. I'm also thinking that we need to abstract the webpack only API's away and your work is the first and most crucial step for this. 

Unfortunately our CI pipeline doesn't run on PR's from forks, but I will have a look at that later. If it would run, you would see that your refactoring caused issues with our "manual test bench" the `example` package. If you run `pnpm example` in the root or `pnpm dev` in the `packages/example` directory you will see an error:

> Error: Error while resolving cross-file selectors in file "/something/next-yak/packages/example/app/page.tsx"
  Caused by: Error parsing file /something/next-yak/packages/example/app/mixins/index.ts: Error parsing exports: exports is not defined

We've started with a similiar migration in order to support vite in our `vite-poc` branch, but that is very much POC and you're API looks better at the first glance. In you're case your especially interested in an esbuild compatible version right? How would that look like if we want to support esbuild with this agnostic cross-file-resolver?

And once again thank you for your help. It's definitely appreciated. I will have a more thorough look when I have more time, but maybe also @jantimon wants to have a look in the mean time?

Mad-Kat:
Thank you for the quick fix.

We've tested everything internally and didn't notice any changes in build time or other errors. I also had a more thorough look into the changes and that also looks good. There are only 2 small things before we can merge:

- Please run `pnpm changeset` in order to create a changelog entry for your change
- We now rely on node only API's for the cross file resolver (which we didn't before) and this breaks the build for the docs (run `pnpm build` in the `packages/docs` folder). There we need to ship everything to the browser so that [the playground](https://yak.js.org/playground) works correctly. Could you have a look at that?

Unfortunately our CI checks are still broken for PR's from forks, otherwise you would have seen these issues. 

I'm positive once these concerns are addressed, we can merge and release a prerelease version.

PaulBlanche:
I have an issue with the last point regarding the `/keyframes` page of the `example` project.

Currently during the parse phase, any exported tag template (`styled`, `css`, `keyframe`, but also any custom tag template) is marked as `tag-template`.

Then during the resolve phase (after `yak-swc`), any `tag-template` matching a generated styled-component comment is resolved to `styled-component` with the generated `className`.

That way only `styled` tag template coming from yak are actually resolved to `styled-component`. So this for example : 
```ts
//foo.js
import styled from "styled-component"

export const foo = styled`...`

//bar.js
import { foo } from "./foo.js"
import { styled } from "next-yak"

const bar = styled.div`
  ${foo} {
    ...
  }
`
```
would error during build with the "unknown type "tag-template"" error message, indicating that the symbol `foo` from the module `./foo.js` is not a yak-compatible styled-component.

---

Ok so this is the behavior in the PR, and it leads to the issue you reported, because of next and its "double build". Next builds once for server-side bundle, once for client-side bundle. During the server-side build, all "use client" module are transformed to throw an error if any exported symbols are used (because it would be invalid to use client module server-side). This transformation removes the comments generated by `yak-swc`. Here it what it look like for the `ClockHands.tsx` module : 
```js
import { registerClientReference } from "react-server-dom-webpack/server";
export const ClockHands = registerClientReference(
function() { throw new Error("Attempted to call ClockHands() from the server but ClockHands is on the client. It's not possible to invoke a client function from the server, it can only be rendered as a Component or passed to props of a Client Component."); },
"/home/whiteshoulders/projects/next-yak/packages/example/app/ClockHands.tsx",
"ClockHands",
);export const ClockHand = registerClientReference(
function() { throw new Error("Attempted to call ClockHand() from the server but ClockHand is on the client. It's not possible to invoke a client function from the server, it can only be rendered as a Component or passed to props of a Client Component."); },
"/home/whiteshoulders/projects/next-yak/packages/example/app/ClockHands.tsx",
"ClockHand",
);
```
So during the parse phase we marked the `ClockHand` as `tag-template` and during the resolve phase, we found no matching comment with a `className`. So the symbol in the server-side bundle stays a `tag-template` that can't be resolved, leading to the error.

---

As a fix, i decided to handle `tag-template` during the resolve phase to resolve them as `styled-component` without `className`. Such a `styled-component` will be interpolated as an empty string in css. For the case of Nextjs, this is not an issue, but this means that yak could "silently" fail in some case.

The most prolematic of them would be in the case of an incomplete migration from "styled-component". Taking again the first example : 
```ts
//foo.js
import styled from "styled-component"

export const foo = styled`
  color: blue;
`

//bar.js
import { foo } from "./foo.js"
import { styled } from "next-yak"

const bar = styled.div`
  ${foo} {
    color: red;
  }
`
```
This would happen for someone migrating from styled-component who forgot to update some modules.

Before the fix, the user would get a build error saying that `foo` is not a yak compatible styled-component (the message could be clearer, but the sentiment is there).

After the fix, the user would get a build success or error depending on the CSS handling tool-chain. For `bar`, the generated css would be something like : 
```
.bar-1234 {
  {
    color: red;
  }
}
```
the usage of `foo` is replaced with an empty string. Some tool-chains might choke on the invalid CSS and give correct diagnostics that might help the user correct the error, but i'd argue that it would be difficult to understand for users, and frustrating to fix.
Worse it the tool-chain is of the "garbage in/garbage out" type and happily processes and outputs malformed CSS.

---

To summarize, i fixed the issue, but doing so introduces the possibility for errors with yak that would be frustratingly difficult to troubleshoot. But i can't see a better fix that would handle the "double build with comment-erasing transform" issue with Next.js without being brittle or over-engineered (like trying to detect if we are in the server-side or client-side build).

Mad-Kat:
@PaulBlanche 

Thank you for the detailed write-up. I think next.js builds actually 3 times (react-server-component, ssr, csr). The build step you describe is the difference between react-server-components, where all client components have to replaced with references/holes that are loaded during SSR and/or on the client. But your conclusion seems to be correct.

How did it work before your PR? Did it just silently fail in the RSC step and populated the right values/class names in the SSR + CSR step? Because [currently the styles work](https://yacijs.vercel.app/keyframes)

image

PaulBlanche:
Sorry i did not make it clear in the previous comment :
- originaly yak resolved any tag template as a styled component, and the classname would be populated if found in comments, leaving some potentially with no classname
- then before the last comment my PR changed the behavior of yak by ensuring that not finding a classname would be a build-time error (i thought it would be better to fail early)
- then the fix reverted back to the current behavior of yak, by resolving any non-resolved tag template as a styled component with no classname

This is why i favored the current fix (even if i'm not happy with it), because it preserves the current behaviour of yak

In the case of next, the original behavior (and behavior of my PR post-fix) made it so in the "server bundle", the generated CSS would be invalid (with empty strings or "undefined" in CSS). But the configuration of webpack in Next.js made it so the CSS is only ever gathered and compiled for the client-side build :
- During the server build, the `./xxx.yak.css!=!...` import would resolve to invalid CSS, but this CSS is never used.
- During the client build the `./xxx.yak.css!=!...` import would resolve to valid CSS (because there the comments from yak-swc are picked up by the loader) which is what webpack uses to output the CSS for the page.

But as i mentioned, using custom tag template (or styled tag template from other packages) would output invalid CSS (currently, and with my PR post fix) : https://yak.js.org/playground?q=H4sIAPx%2Bh2gCA51TzWrbQBDes55icJpig%2FXj2KFBcgKl0GMvzSWUgtfSSlq8spTVxnVq3BJa%2Bkugl0JPTZ9FT9bZlZTYxOQQg%2BT52f3mm29GHqGk%2BocvZpHq6ylXgvXhZZ6T6pvjkuqmE1gWEaS6Jh9gRaovXQvwN37B5orJE%2BOgG%2FFFYwKQ6vutndGlTao%2FMPK8Yhk04cm6MdrrCMCz5NYBKGV43CG8sxHaRN1EHnjefgu8BQ7gbuCbxk5esaWyz%2Bhs7NZ%2BS99F%2FrUzdu866wXWGpsP0XyyQknWsDJnSPXXh6m4YLru2poEltbwBl8xHj9XKcsYHJvDEZUzHzqpyoSjbXja6WNY8CRVbdw4JrHWWlfXcZ53S0UVD0v%2FlGWFoIq9VpLPk%2FK5lPSyD47jUJmUPtD55Zu3PVPKcD1HmlgaESYYMkQliwJronOmZ%2FIxHdS5z3AoWaZ7IAs9IGOtIDT9a0dptab50sbamQ%2F6ja5Kt3IsSpgPIS2AiiKlU4a0sQ1d4MqHPa%2FB1Y6kEacCp%2FbDyBhyGQoGMZWoWKnsMJdzJoEqUHkBgsVKSwUGA7z9xhkOh2bm9Xw08i%2FDJeZCQ2PD5H1d%2F12bKpXMZ0gSlYBBsazl0jc%2F4fPzrGuPMFjvEKl%2B%2B4AuGNZ67maaZnbt%2BJtuyGx0GLGkD3vRIBw889Bg0ZF3pI060gNbM3XhYGNHkRRet0PBC5QU6d0l7kfrXX58ufvi8DlXOITtdCuQblyr9wDP3TEEXlDZtXUNO47w76oRVDCF22SXBQ1xf33wHO%2BgXjpc111wO%2BpOAvy0LvCJ%2FgO3WaDlrQQAAA%3D%3D
The typechecking in the `styled` call in `Center` show that there is an error, but what if the user does not typecheck ?

Mad-Kat:
I get it. Thank you for the very good explanation (and playground to see it in action). I recognize the issues with these silent failures (#368, #369, #370, #373), where it would have been helpful to get an early error feedback. We thought that the best bet currently would be to have the safety with TypeScript, but this just handles the basic cases. 

I think this case will get easier once we have access to the unified graph approach of Turbopack and Rollup/Rolldown, but I'm not yet sure, as this is a complex problem. Maybe we could also instrument the client reference transformer to inject our comments, but I'm not sure if that is even possible. 

@jantimon what are your thoughts on this issue?

packages/next-yak/cross-file-resolver/resolveCrossFileConstant.ts
```diff
@@ -0,0 +1,677 @@
...
+      // tag-template would be replaced with `""` (not very user friendly)
+      case "tag-template": {
+        return {
+          type: "styled-component",
+          from: [filePath],
+          source: filePath,
+          name: specifiers[specifiers.length - 1],
+          value: undefined,
+        };
+      }
```
Mad-Kat, line 517: This isn't always a `styled-component`. It can also be a mixin dependent on the importKind

PaulBlanche: Yes, but currently it is already handled that way: https://github.com/DigitecGalaxus/next-yak/blob/215820b00834c26552d9f5f11a04f418bd144863/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts#L507

When parsing an expression, any tag template is marked as styled component (even custom tags or mixins). It is latter, when matching mixins/styled components extracted from comments with symbols in exports that the distinction mixin/styled-component is done : https://github.com/DigitecGalaxus/next-yak/blob/215820b00834c26552d9f5f11a04f418bd144863/packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts#L257

To avoid confusion instead of marking tag template as styled component during parsing, i mark them as tag template, and later during the resolve phase i do the matching between mixins/styled components extracted from comments and symbols in export. Any remaining tag-template is handled as it would have been previously : like a styled component with no classname.

I agree this feels wrong, and my refactoring brings light to this "incorrectness" (which was a bit hidden previously). Maybe i could introduce another kind of symbol like "non-resolved-tag" instead of marking them as styled-component (it would feel less wrong). Then when interpolating in the css, those "non-resolved-tag" would be replaced by empty strings (like style-template with no classnames currently).

Mad-Kat (approved):

Mad-Kat:
Changes were deployed with version: [6.1.0-next.0](https://www.npmjs.com/package/next-yak/v/6.1.0-next.0)

PR #390, opened 2025-07-22
Add support for default exported components

rvetere:
Implement support for default exported components, that should fix https://github.com/DigitecGalaxus/next-yak/issues/370

### DIFF
--- a/packages/cross-file-tests/__tests__/fixtures/constant/output/index.tsx
+++ b/packages/cross-file-tests/__tests__/fixtures/constant/output/index.tsx
@@ -2,7 +2,7 @@ import { styled, css } from "next-yak/internal";
 import { colors, negative, siteMaxWidth } from "./constants";
 import * as __yak from "next-yak/internal";
 import "./index.yak.module.css!=!./index?./index.yak.module.css";
-export var Button = /*YAK EXPORTED STYLED:Button:index_Button_wIxfWW*//*YAK Extracted CSS:
+export var Button = /*YAK EXPORTED STYLED:Button:index_Button_wIxfWW*/ /*YAK Extracted CSS:
 :global(.index_Button_wIxfWW) {
   color: red;
   height: --yak-css-import: url("./constants:siteMaxWidth",mixin)px;
--- a/packages/cross-file-tests/__tests__/fixtures/namespaceConstant/output/index.tsx
+++ b/packages/cross-file-tests/__tests__/fixtures/namespaceConstant/output/index.tsx
@@ -2,7 +2,7 @@ import { styled } from "next-yak/internal";
 import { colors } from "./constants";
 import * as __yak from "next-yak/internal";
 import "./index.yak.module.css!=!./index?./index.yak.module.css";
-export var Button = /*YAK EXPORTED STYLED:Button:index_Button_bAtfrp*//*YAK Extracted CSS:
+export var Button = /*YAK EXPORTED STYLED:Button:index_Button_bAtfrp*/ /*YAK Extracted CSS:
 :global(.index_Button_bAtfrp) {
   color: --yak-css-import: url("./constants:colors:primary",mixin);
   background-color: --yak-css-import: url("./constants:colors:secondary",mixin);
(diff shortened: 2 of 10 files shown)


packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -224,24 +232,40 @@ impl YakTransform for TransformCssMixin {
...
+        // For default exports, separate the comments:
+        // - Export comment goes on export default statement
+        // - No CSS comment for mixins (CSS is embedded in the mixin comment)
+        let export_comment = Some(format!("{}{}", YAK_EXPORTED_MIXIN_PREFIX, export_name));
+        (export_comment, None)
+      } else {
+        // For named exports, keep comment at variable declaration
+        let full_comment = Some(format!("{}{}", YAK_EXPORTED_MIXIN_PREFIX, export_name));
+        (full_comment, None)
+      }
```
Mad-Kat, line 260: The two branches aren't different

packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -214,7 +222,7 @@ impl YakTransform for TransformCssMixin {
         .into(),
       );
     }
-    let css_prefix = if self.is_within_jsx_attribute {
+    let (css_prefix, css_only_comment) = if self.is_within_jsx_attribute {
```
Mad-Kat, line 225: I think the tuple doesn't make much sense here, as there is only one case where `css_only_comment` is set.

packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -479,19 +506,41 @@ impl YakTransform for TransformStyled {
...
-    let css_prefix = if self.is_exported {
-      Some(format!(
-        "YAK EXPORTED STYLED:{}:{}*//*YAK Extracted CSS:",
-        self.declaration_name.to_readable_string(),
-        self.class_name
-      ))
+    let (css_prefix, css_only_comment) = if self.is_exported {
+      // For default exports (both inline and separated), always use "default"
+      // For named exports, use the variable name
+      let declaration_name_str = self.declaration_name.to_readable_string();
```
Mad-Kat, line 512: this could be inlined into the `else` case

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -135,6 +137,8 @@ where
   display_names: bool,
   /// Transpilation mode to determine how to transpile the code
   transpilation_mode: TranspilationMode,
+  /// Storage for default export styled components that need special comment handling
+  default_export_styled_components: FxHashMap<String, String>,
```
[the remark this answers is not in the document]
Mad-Kat: I think `current_default_export` is also not needed

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -620,6 +647,44 @@ where
     self.current_exported = false;
   }
 
+  /// To store the current export state for default exports  
+  /// e.g. export default styled.button`color: red;`
+  fn visit_mut_export_default_decl(&mut self, n: &mut ExportDefaultDecl) {
```
Mad-Kat, line 652: I think we don't need this case, as the Variable visitor already knows if the variable is exported

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -803,13 +868,20 @@ where
...
-      "styled" if is_top_level => Box::new(TransformStyled::new(
-        &mut self.naming_convention,
-        current_variable_id.clone(),
-        self.display_names,
-        self.current_exported,
-        self.transpilation_mode,
-      )),
+      "styled" if is_top_level => {
+        let is_exported =
+          self.current_exported || self.variables.is_default_exported(&current_variable_id.id);
```
Mad-Kat, line 873: When is the first one false and the second one true? This case can't happen

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -824,13 +896,20 @@ where
...
-      "css" if is_top_level => Box::new(TransformCssMixin::new(
-        &mut self.naming_convention,
-        current_variable_id.clone(),
-        self.current_exported,
-        self.inside_element_with_css_attribute,
-        self.transpilation_mode,
-      )),
+      "css" if is_top_level => {
+        let is_exported =
+          self.current_exported || self.variables.is_default_exported(&current_variable_id.id);
```
Mad-Kat, line 901: same here

Mad-Kat (changes requested):

PR #422, opened 2025-10-26
Run `parseYakCssImport` in parallel

chwoerz:
Currently it looks like `parseYakCssImport` runs sequentially because of the `for of` and `await`. I changed that and made it so it returns a promise and resolved it using `Promise.all`. 

The downside is, that I had to change it from an iterator (which matchAll) returns, to an array. If that is not wanted, just decline the PR.


Mad-Kat:
Hey @chwoerz 

Thank you so much for this nice PR. What is the motivation behind it? Did you read the code and noticed some performance improvements or did you have a problem with the slowness of yak?

chwoerz:
@Mad-Kat I just saw the repository mentioned on galaxus and gave it a read. 
I'm not using it myself but I'm sure it's already fast enough. 😁

I would have some small refactoring proposals I would also create a PR for, if it's okay. Mainly reusing code and removing duplications. 

Btw: I could not run the benchmark tests locally as it stated that letter.js is not available. It looks like the js files didn't get created. I read the CONTRIBUTIONS.md but still I could not get them to work. Could you maybe provide a small step-by-step guide how to be able to run the benchmarks? (Including pre-steps)

Issue #481, opened 2026-02-07
Exported variable '...' has or is using name 'YakComponent' from external module 'node_modules/long_path' but cannot be named


nikelborm:
I'm getting this ts error. I'm using the latest Bun as a package manager.
```plaintext
Exported variable 'SelectTrigger' has or is using name 'YakComponent' from external module "/home/nikel/projects/effect-garden/node_modules/.bun/next-yak@9.1.0+4cb091f5c2bf0ada/node_modules/next-yak/dist/index" but cannot be named.ts(4023)
```

```ts
import { css, styled } from 'next-yak'

export const SelectPositioner = styled(BaseSelect.Positioner)`
  outline: none;
  z-index: 1;
  user-select: none;
`
```

nikelborm:
It also seems very strange to me that almost all of the types and interfaces inside `/dist/index.ts` are not exported:
```ts
interface YakTheme {
}

type ComponentStyles<TProps> = (props: TProps) => {
    className: string;
    style?: {
        [key: string]: string;
    };
};
type CSSInterpolation<TProps> = string | number | undefined | null | false | ComponentStyles<TProps> | {
    [yakComponentSymbol]: any;
} | ((props: TProps) => CSSInterpolation<TProps>);

interface Styled extends MappedHtmlTags, StyledFn {
}
interface StyledFn {
    <TProps extends object = React.DOMAttributes<Element> & React.RefAttributes<Element>>(Component: HtmlTags | React.FunctionComponent<TProps> | CustomWebComponentTag): LiteralWithAttrs<TProps>;
}
interface YakComponent<T> extends React.FunctionComponent<T> {
    [yakComponentSymbol]: [unknown, unknown, unknown];
}
interface LiteralWithAttrs<T extends object> extends StyledLiteral<T> {
    attrs: <TAttrsIn extends object = {}, TAttrsOut extends AttrsMerged<T, TAttrsIn> = AttrsMerged<T, TAttrsIn>>(attrs: Attrs<T, TAttrsIn, TAttrsOut>) => StyledLiteral<Substitute<T, TAttrsIn>>;
}
interface StyledLiteral<T> {
    <TCSSProps>(styles: TemplateStringsArray, ...values: Array<CSSInterpolation<T & NoInfer<TCSSProps> & {
        theme: YakTheme;
    }>>): YakComponent<TCSSProps & T>;
}
[...]

PR #483, opened 2026-02-08
Fix TS4023 when re-exporting styled components

Mad-Kat:
Export `YakComponent` type so that re-exporting styled components works when declartion emit is active.

Similar issue to [the one in styled-components](https://github.com/styled-components/styled-components/issues/4299)

This closes: #481

### DIFF
--- a/.changeset/open-pots-argue.md
+++ b/.changeset/open-pots-argue.md
@@ -1,0 +1,5 @@
+---
+"next-yak": patch
+---
+
+Export the YakComponent type to fix TS4023 errors when re-exporting styled components with declaration emit enabled


packages/next-yak/runtime/index.ts
```diff
@@ -27,7 +27,7 @@
 export { useTheme, YakThemeProvider } from "next-yak/context";
 export type { YakTheme } from "./context/index.d.ts";
 
-export type { GenericYakComponentOf } from "./publicStyledApi.ts";
+export type { GenericYakComponentOf, YakComponent } from "./publicStyledApi.ts";
```
jantimon: is `packages/next-yak/runtime/index.ts` correct here to export it from the next-yak npm package?

PR #493, opened 2026-02-27
Add basePath option to vite plugin

Mad-Kat:
This PR adds a `basePath` option to the vite plugin that is helpful in a monorepo setup where files can be outside of the current vite root.


packages/next-yak/loaders/vite-plugin.ts
```diff
@@ -217,10 +235,23 @@ export async function viteYak(
       },
       async handler(code, id) {
         try {
+          const filePath = id.split("?")[0];
+          if (!hasWarnedAboutBasePath && !userOptions.basePath) {
```
jantimon: if `relPath.startsWith("..")` we should probably still warn (probably the user misconfigured yak)
```suggestion
          if (!hasWarnedAboutBasePath) {
```

what do you think?

PR #515, opened 2026-03-19
Fix CSS url() asset references

Mad-Kat:
Fixes [#513](https://github.com/DigitecGalaxus/next-yak/issues/513)

**Root cause:** The `\0` prefix on virtual module IDs prevents Vite's CSS pipeline from determining the correct directory for `url()` resolution.

**Fix:** Drop the `\0` prefix and use a `.yak.css` suffix instead. `resolveId` now returns `/abs/path/Component.tsx.yak.css`, giving Vite the correct directory context to resolve relative `url()` paths. This follows the same convention as [vanilla-extract](https://github.com/vanilla-extract-css/vanilla-extract/blob/9e7a583c935/packages/vite-plugin/src/index.ts#L282), [Linaria](https://github.com/Anber/wyw-in-js/blob/86dae418a0e/packages/vite/src/index.ts#L576-L577), and [Pigment CSS](https://github.com/mui/pigment-css/blob/7d763a05bac/packages/pigment-css-vite-plugin/src/vite-plugin.ts#L103-L105). Vite [docs ](https://vite.dev/guide/api-plugin.html#virtual-modules-convention) confirms that `\0` isn't needed for file-derived modules.

**Cross-file url() references:** A second issue exists with `url()` paths inside cross-file mixins. When a mixin is inlined into a consuming component, relative `url()` paths still point relative to the mixin's original directory, not the consuming file's directory. To fix this, relative `url()` paths are now rewritten during cross-file resolution so they resolve correctly from the consumer's location.


jantimon (dismissed):

jantimon:
nextjs with webpack fails because of a bug in mini-css-extract plugin which was resolved by @Mad-Kat in https://github.com/webpack/mini-css-extract-plugin/issues/1161 and merged - so we have to wait for a new next.js release with mini-css-extract-plugin 2.10.2

PR #537, opened 2026-04-29
Harden GitHub Actions workflows

Mad-Kat:
Static analysis with [zizmor](https://docs.zizmor.sh) flagged 66 findings across the workflow suite. This PR addresses all of them and adds zizmor as a CI gate to keep things clean going forward.


jantimon (approved):

.github/workflows/zizmor.yml
```diff
@@ -0,0 +1,37 @@
+name: Zizmor
```
jantimon: lets find a name a junior dev understands

PR #552, opened 2026-05-13
Publish canary snapshots on every push to main

Mad-Kat:
Adds a `prerelease` job to the release workflow. Every push to `main` with pending changesets publishes a snapshot (e.g. `9.4.2-canary-20260513120000`) to npm under the `canary` dist-tag, plus a per-package GitHub prerelease.

```
npm i next-yak@canary
```

Mechanism: Changesets [snapshot mode](https://github.com/changesets/changesets/blob/main/docs/snapshot-releases.md), mirroring styled-components' [release.yml](https://github.com/styled-components/styled-components/blob/main/.github/workflows/release.yml).


Mad-Kat:
as discussed. This isn't needed at the moment

PR #558, opened 2026-06-10
perf(runtime): string className collector, chain flattening, static fast path — every SSR bench cell now beats styled-components

jantimon:
This branch rewrites the runtime render path. No API changes, rendered output identical, all runtime tests green — your existing components just render faster, with nothing to change on your side.

**On a real page, against the version you're running today**

A 400-tile product listing (responsive grid, sale badges, wishlist toggles, dynamic rating bars, `@container` queries, focus-visible rings — a real page, not a toy button), `perf` vs the released `next-yak 9.6.0`:

yak-perf-benches

- **SSR render throughput:** 75,925 → **110,092 renders/sec** (+45%)
- **SSR under concurrent load:** 133 → **170 req/sec** (+28%)
- **Server CPU per render:** 5.29 → **3.71 ms** (−30%) — the styling-runtime slice (orange) all but disappears
- **Client hydration:** 18.5 → **16.1 ms** (−13%), **cold mount:** 42.8 → **39.6 ms** (−7%)

That last pair matters: the win isn't only SSR throughput, it's less main-thread work in the browser too. Interaction (in-place re-render) is a tie — that path is dominated by browser layout, not styling.

**Where it comes from**

Four changes did the work. Here's each one, the code in *your* app it speeds up, and how much each synthetic pattern moved against the same code on `main`:

yak-perf-branch

**1. Nested `styled(...)` chains — the biggest win (+430%)**

Extending a styled component, then extending that again:

```tsx
const Button = styled.button`
  color: white;
`;
const Primary = styled(Button)`
  background: blue;
`;
const CallToAction = styled(Primary)`
  font-size: 2rem;
`;
```

Rendering one `<CallToAction />` used to walk every level of the chain at render time — each wrapper calling its parent, redoing the theme read, prop clones and class merging the outer wrapper had already done. The chain now flattens when the component is *defined*: one wrapper renders the final `<button>` directly, no matter how deep the chain goes.

**2. Fully static components take a fast path — the common case (+100% and up)**

Any component with no `${...}` interpolation and no dynamic attrs:

```tsx
const Card = styled.div`
  padding: 16px;
  border-radius: 8px;
`;
```

This is most components in most apps. It used to run the full dynamic machinery on every render — build a Set of class names, subscribe to theme context, clone props and style, set internal markers — none of which it needs. It now early-returns: strip `$`-props, append the class string, render. It also stops subscribing to theme, so on the client these components no longer re-render on a theme switch (they never read the theme).

**3. className building is a string append, not a `Set` round-trip (drives the tree cases)**

Every styled render used to do this, per element:

```tsx
const set = new Set(className?.split(" "));
set.add(ownClassName);
className = Array.from(set).join(" ");
```

In the nested benchmark that round-trip alone was 53% of all process CPU. The compiler already emits a distinct class per component, so the `Set`'s dedup bought nothing in the hot path. It's now a plain string append (the `has`/`delete` contract is kept for atoms and css-prop merging). That's why the deep and wide trees moved most — the cost was paid per class name, per element.

**4. Constant `.attrs()` also takes the fast path (+46%)**

```tsx
const SubmitButton = styled.button.attrs({ type: "submit" })`
  color: white;
`;
```

`.attrs()` with a constant object wraps it in a function internally, which made every attrs component miss the static fast path above. It now recognizes the constant and routes through the same fast path.

**The cost: +170 bytes**

yak-bundle-size

Runtime size (`internal.js`, minified + brotli) across the branch: +170 B total, 2517 → 2687. The string collector from the className change above is the biggest single piece at +83 B. On the shop page above that shows up as ~200 B more gzipped JS — the trade is a handful of bytes for 1.3–5× render throughput.

**Safety**

- No public API changes — every exported type is byte-identical.
- Rendered HTML is identical across the benchmark suite and the e2e tests.
- All runtime tests green.
- Verified commit-by-commit against `main`: nothing regressed. The only non-win is the css-prop case, flat within measurement noise

### DIFF
--- a/.changeset/drop-classname-dedup.md
+++ b/.changeset/drop-classname-dedup.md
@@ -1,0 +1,5 @@
+---
+"next-yak": minor
+---
+
+The runtime no longer deduplicates class names, so the same class applied twice — e.g. an `atoms()` utility that is already on the element, or a shared `css` block toggled by two conditions — now appears twice in the class attribute. This is rendering-neutral, as CSS applies a class once however often it is listed.
--- a/packages/next-yak/runtime/styled.tsx
+++ b/packages/next-yak/runtime/styled.tsx
@@ -224,13 +224,15 @@ const removeNonDomProperties = <T extends Record<string, unknown>>(obj: T): T =>
   return result;
 };
 
-// util function to merge class names, as they are concatenated with a space
-const mergeClassNames = (a?: string, b?: string) => {
-  if (!a && !b) return undefined;
-  if (!a) return b;
-  if (!b) return a;
-  return a + " " + b;
-};
+/**
+ * util function to merge class names, as they are concatenated with a space
+ * e.g.:\
+ * `("a", "b")` → `"a b"`\
+ * `("a", undefined)` → `"a"`\
+ * `(undefined, "b")` → `"b"`\
+ * `(undefined, undefined)` → `undefined`
+ */
+const mergeClassNames = (a?: string, b?: string) => (a && b ? a + " " + b : a || b || undefined);
 
 /**
  * merge props and processed props (including class names and styles)
--- a/packages/next-yak/runtime/styled.tsx
+++ b/packages/next-yak/runtime/styled.tsx
@@ -47,8 +47,10 @@ const styledFactory: StyledFn = (Component) =>
  */
 export const styled = styledFactory as Styled;
 
-// Real internal shape of the yakComponentSymbol tuple. Public YakComponent keeps it
-// opaque ([unknown, ...])
+/**
+ * Real shape of the yakComponentSymbol tuple, which the public `YakComponent`
+ * keeps opaque as `[unknown, ...]`
+ */
 type YakComponentInternals = [
   self: React.FunctionComponent,
   attrsFn: AttrsFunction<any, any, any> | undefined,
@@ -131,7 +133,6 @@ const yakStyled: StyledInternal = (Component, attrs) => {
       // attrs functions and dynamic style functions receive the theme —
       // fully static components take the fast path above and never read the
       // theme context
-      //
       const theme = useTheme();
 
       // The first components which is not wrapped in a yak component will execute all attrs functions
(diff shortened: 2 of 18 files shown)


.changeset/drop-classname-dedup.md
Mad-Kat, line 1: We should merge the changesets. This is only one feature

benchmarks/bench/cross-request-cache/gen.ts
```diff
@@ -114,3 +114,83 @@ export const CrossRequestCache${
...
+}
+
+const Box: FunctionComponent<BoxProps> = ({ $color, $layout, $outer, $fixed }) => (
+  <div
+    className={$fixed ? "view box boxFixed" : "view box"}
+    style={{
+      '--fd': $layout === 'column' ? 'column' : 'row',
+      '--p': $outer ? '4px' : '0',
+      '--bg': getColor($color ?? -1),
+    } as React.CSSProperties}
```
Mad-Kat: Would you write it this way? This looks very similar to compiled yak code 😉

benchmarks/experiments/e1-custom-prop-invalidation.html
Mad-Kat, line 1: Should we keep this checked in? For what is it used?

packages/next-yak/runtime/styled.tsx
```diff
@@ -58,15 +52,26 @@ const yakStyled: StyledInternal = (Component, attrs) => {
...
         RuntimeStyleProcessor<unknown>,
+        React.FunctionComponent | string,
       ])
     : [];
 
+  // the ultimate render target of the whole styled(styled(...)) chain:
+  // attrs and style processors are already merged at construction time, so
+  // a chain of N levels renders the target directly in ONE wrapper instead
+  // of re-entering every parent wrapper per element per render
+  // (the re-entry was the structural ×N multiplier in EXP-20260609-02)
```
Mad-Kat:
```suggestion
```

packages/next-yak/runtime/styled.tsx
```diff
@@ -80,20 +85,48 @@ const yakStyled: StyledInternal = (Component, attrs) => {
...
+      // attrs functions and dynamic style functions receive the theme —
+      // fully static components take the fast path above and never read the
+      // theme context
       //
-      // const Button = styled.button`${({ theme }) => css`color: ${theme.color};`}`
-      //       ^ must be have access to theme, so we call useTheme()
-      const theme = mergedAttrsFn || runtimeStylesFn.length ? useTheme() : noTheme;
+      // (this previously gated on `runtimeStylesFn.length` — the function
+      // ARITY, which is always ≥2 — so useTheme() ran for every component
+      // including fully static ones; see EXP-20260609-02)
```
Mad-Kat:
```suggestion
```

Mad-Kat (approved):

Mad-Kat (approved):

PR #559, opened 2026-06-10
build(swc): fix wasm plugin linking on rustc 1.96 + upgrade swc_core 56 → 68

jantimon:
the benchmarks workflow broke for every branch (including unmodified main, which it builds as a baseline) when the GitHub runner image moved from rustc 1.94.1 to 1.96.0. the failure:

```
rust-lld: error: undefined symbol: __set_transform_result
rust-lld: error: undefined symbol: __emit_diagnostics
```

those are SWC *host imports* — functions the swc plugin runner provides at runtime. they're supposed to stay undefined at link time and become wasm imports. [rust 1.96 removed the implicit `--allow-undefined`](https://blog.rust-lang.org/2026/04/04/changes-to-webassembly-targets-and-handling-undefined-symbols/) that made that work, and swc_core (even latest 68.0.6) doesn't yet declare its imports with `#[link(wasm_import_module)]`, so the link now hard-fails on any current toolchain. nothing in our code changed — the toolchain moved under us (we had no pin, so CI silently follows the runner image).

three things in here:

1. **`-Clink-arg=--allow-undefined` in `build:yak`** — the explicit opt-back-in the rust blog recommends until swc_core ships proper import declarations. no-op on older rustc (it just re-adds what used to be default). this is the actual fix.
2. **swc_core `=56.0.0` → `=68.0.6`** — we were 12 majors behind; the separately-pinned swc crates move with it (`swc_ecma_transforms_testing` =48.0.0, `swc_ecma_parser` =41.0.1, `swc_config` =5.0.0). pleasantly boring: zero code changes, all 332 cargo tests green with unchanged snapshots. doesn't fix the linking issue by itself but we shouldn't stay pinned to a year-old ecosystem, and the planned `@property` compiler work wants a current swc_core anyway.
3. **`rust-toolchain.toml` pinning 1.96.0** (+ wasm targets, clippy, rustfmt) — so the next runner-image rustc bump can't silently change how the plugin builds again. rustup picks it up locally too, so CI and dev machines build identically.

verified locally:
- wasm32-wasip1 and wasm32-unknown-unknown (playground) build on rustc **1.96.0** — the exact version that broke CI
- the built plugin runs under **@swc/core 1.15.32** (bench fixture generation through the real wasm produces correct output, so the plugin ABI is compatible with our npm-side swc)
- `cargo test` 332 green / snapshots unchanged, `cargo fmt --check` clean, clippy at parity with main (warnings only)
- full JS test suite green (next-yak vitest + typechecks + example tests)

note: #558 carries a stopgap `rust-toolchain.toml` pinned to 1.94.1 (the last green runner version) so its CI runs now — whichever PR lands second resolves the trivial conflict to this branch's version (1.96.0 + the link-arg).

**added after the first round of checks:**
- changeset (yak-swc patch) — the published wasm artifact changes with this
- **wasm-pack 0.14.0 → 0.15.0**: turns out the "Docs PR preview" workflow has been red on *every* branch since May 18 — wasm-pack 0.14.0 dies with `Error fetching release: 404` downloading its tooling. reproduced locally, 0.15.0 fixes it. so this PR un-breaks the docs preview too.
- **@playwright/test 1.59.1 → 1.60.0**: E2E CI has also been dead since May 18 — every run times out at 30min because `playwright install chromium` hangs after the download hits 100%. known upstream bug ([yauzl extraction regression on Node 24.16+](https://github.com/microsoft/playwright/issues/40747), fixed in 1.60.0); the e2e workflow's unpinned `node-version: 24` floated onto 24.16.0 with the same runner-image update that brought rustc 1.96. verified locally: all 5 bundler lanes green, 27/27 each.

so in total this PR revives three CI pipelines that the May/June runner-image update killed (benchmarks via rustc 1.96 link fix, docs preview via wasm-pack, e2e via playwright) — none of which had pinned versions, so they all drifted.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.changeset/swc-core-68-rustc-196.md
+++ b/.changeset/swc-core-68-rustc-196.md
@@ -1,0 +1,7 @@
+---
+"yak-swc": patch
+---
+
+Fix building the SWC plugin on Rust 1.96+ and upgrade swc_core from 56.0.0 to 68.0.6.
+
+Rust 1.96 removed the implicit `--allow-undefined` for wasm targets, which broke linking the plugin (`undefined symbol: __set_transform_result`). The wasm build now passes the link-arg explicitly, the Rust toolchain is pinned via `rust-toolchain.toml`, and swc_core (plus the separately pinned swc crates) moved to the current major. No behavior changes — all transform snapshots are unchanged.
--- a/benchmarks/bench/.gitignore
+++ b/benchmarks/bench/.gitignore
@@ -1,2 +1,4 @@
 # Generated benchmark files
 generated/
+dist-profile/
+profiles/
--- a/packages/yak-swc/Cargo.lock
+++ b/packages/yak-swc/Cargo.lock
@@ -147,9 +147,9 @@ dependencies = [
 
 [[package]]
 name = "binding_macros"
-version = "54.0.0"
+version = "66.0.0"
 source = "registry+https://github.com/rust-lang/crates.io-index"
-checksum = "674b724803669fe7b32b82d87b66d93f028bf686c4c14f2ae5feef1457cac51c"
+checksum = "36d411e8a776673d0b5f4f07ce0bd25b0cb231d1fb17c4fa13f47abb3aebb0e4"
 dependencies = [
  "anyhow",
  "console_error_panic_hook",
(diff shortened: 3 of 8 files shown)


Mad-Kat:
Did you test that this `swc_core` is compatible with the next.js version? I know since they introduced non-breaking updates it should be fine, but double checking wouldn't hurt

Mad-Kat (approved):

PR #508, opened 2026-03-15
add rspack support

jantimon:
adds rspack using option B from #507

PR #562, opened 2026-06-23
Add rsbuild 2 support

### LINKED ISSUE #507: Rspack Support for next-yak
[Rspack joins the Next.js ecosystem](https://rspack.rs/blog/rspack-next-partner) so I guess we should add support for rspack

To understand the options we have for rspack, it helps to know how the other bundlers solve the same problem. The core challenge is always the same: the yak SWC plugin transforms a `.tsx` file and embeds CSS as comments in the JS output. We then need to extract that CSS into a separate module. The tricky part is that the class name hashes must match exactly between the JS runtime and the generated CSS, so ideally both sides come from the same SWC transformation.

**Webpack** solves this with `this.loadModule()`. The CSS loader calls `this.loadModule(this.resourcePath)` which re-processes the file through webpack's loader pipeline (including SWC). It returns the transformed source as a string, we extract the CSS from it. One SWC pass, hashes always match.

**Turbopack** takes a different route. The SWC plugin is configured to emit CSS directly as base64 data URL imports (`import "data:text/css;base64,..."`) in the transformed JS. No separate CSS loader needed, turbopack picks up the data URLs natively.

**Vite** has neither `loadModule` nor a loader pipeline to hook into. So the vite plugin reads the source file and calls `swcTransform()` directly a second time with the same config, then extracts CSS from that. It works, but SWC runs twice per file and hash consistency depends on both calls using identical config.

## The problem with rspack

Rspack is mostly webpack-compatible but is missing the two APIs we rely on:

- `this.loadModule()` is not implemented. Calling it throws `TypeError: this.loadModule is not a function`
- Data URL imports (`import "data:text/css;base64,..."`) are not supported out of the box

Rspack does have `this.importModule()`, but that one compiles and evaluates the module. Since we're dealing with React component files, that would try to execute React code at build time which obviously doesn't work. You could work around that with an inline raw loader (see Option A), but `importModule` also recompiles the module in Node.js mode to make it evaluatable, so it's a separate compilation from rspack's normal browser-targeted build anyway

Also Source maps are passed to loaders as JSON strings instead of objects, which breaks SWC. Fixable by parsing/serializing at the boundaries, but something to be aware of

So we need a different approach. We explored two options:

## Option A: `importModule` + inline raw loader

The idea here is to work around the `importModule` evaluation problem. If we pipe the file through a tiny raw loader first, the module's export becomes just a string instead of React code.

The raw loader is literally 3 lines:

```ts
export default function rawLoader(source: string) {
  return `export default ${JSON.stringify(source)}`;
}
```

Then in the CSS extraction loader:

```ts
const { default: source } = await this.importModule(
  `${pathToRawLoader}!${this.resourcePath}`
);
const css = extractCss(source, transpilationMode);
```

What happens under the hood:

1. `importModule` triggers a separate compilation of the `.tsx` file in Node.js mode
2. Rspack's loaders (including SWC with the yak plugin) run as part of that compilation
3. The inline raw loader wraps the output as `export default "<source>"`
4. `importModule` evaluates the module and returns the string

Important caveat: since `importModule` compiles in Node.js mode, this is a separate SWC pass from the browser-targeted one. The yak plugin config should be the same, so hashes should match, but it's not guaranteed by the pipeline itself. This is the same situation as Vite

## Option B: Reuse the turbo-loader + `experiments.css`

This approach was implemented and tested in `yak-worktree-2`. Instead of replicating the webpack flow, it reuses the existing turbo-loader (originally built for turbopack) and enables rspack's native CSS support to handle the data URL imports.

How it works:

1. Detect rspack via the `NEXT_RSPACK` env var (set by `withRspack()`)
2. Register the turbo-loader as an `enforce: "pre"` rule for source files
3. The turbo-loader runs SWC directly with the yak plugin, extracts CSS, and outputs JS with `import "data:text/css;base64,..."`
4. Enable `experiments.css = true` and add a rule for `{ scheme: "data", mimetype: "text/css", type: "css/auto" }` so rspack handles those data URL imports natively

```
source.tsx -> turbo-loader (runs SWC + extracts CSS)
                   |
    JS with data:text/css;base64 import
                   |
    rspack's built-in SWC (handles remaining JSX)
                   |
    rspack's native CSS pipeline (experiments.css)
```

A few things to note:

- Config wrapper order matters: `withYak(withRspack(nextConfig))` because `withRspack` must run first to clean up the `TURBOPACK` env var before `withYak` checks it
- The turbo-loader preserves JSX (`react.runtime: "preserve"`), so rspack's built-in SWC still handles JSX->JS compilation after
- Source maps needed some fixing because rspack passes string source maps where SWC expects objects

## Comparison

Both options run SWC twice per file and both rely on keeping the yak plugin config in sync between the two passes for hash consistency. Neither can reuse rspack's browser-targeted SWC pass directly (unlike webpack's `loadModule` which does exactly that).

| | Option A: `importModule` + raw loader | Option B: turbo-loader + `experiments.css` |
|---|---|---|
| SWC runs per file | Twice (Node.js recompilation) | Twice (turbo-loader + rspack's built-in) |
| Hash consistency | Depends on keeping SWC config in sync | Depends on keeping SWC config in sync |
| New code needed | ~3-line raw loader + rspack CSS loader | Only `addYakRspack()` in withYak |
| CSS pipeline | Next.js's CSS pipeline | rspack's native CSS pipeline |

jantimon:
Adds a next-yak/rsbuild plugin for rsbuild 2 / rspack 2. Runs yak-swc as a pre loader and inlines CSS as a data URL, same as the turbopack integration. Drops into the plugins array next to pluginReact, so rsbuild-based setups like TanStack Router's "with rspack" guide work too.

Includes an rsbuild example + e2e coverage. Quirk: rsbuild does a full page reload on yak.ts syntax-error recovery instead of a hot update, so that one assertion is skipped for rsbuild.

```
Total
┌────────────────────────┬───────────┬────────┐
│ Bundler                │ Duration  │ Passed │
├────────────────────────┼───────────┼────────┤
│ next-app-turbopack     │ 7m 53.3s  │ 27/27  │
│ next-app-webpack       │ 13m 20.9s │ 27/27  │
│ next-pages-turbopack   │ 6m 24.5s  │ 27/27  │
│ next-pages-webpack     │ 7m 20.7s  │ 27/27  │
│ rsbuild                │ 4m 5.1s   │ 27/27  │
│ tanstack-start-rsbuild │ 4m 39.0s  │ 27/27  │
│ vite                   │ 4m 14.5s  │ 27/27  │
└────────────────────────┴───────────┴────────┘
```

fixes #507

### DIFF
--- a/e2e/cases/hmr-yak-syntax-error-recovery/index.test.ts
+++ b/e2e/cases/hmr-yak-syntax-error-recovery/index.test.ts
@@ -36,8 +36,6 @@ test(
 
     /** rsbuild recovers from the error via a full page reload instead of a hot update */
     const bundlerSupportsHmrErrorRecovery = testEnv.bundlerDirName !== "rsbuild";
-    expect(Boolean(await page.evaluate(() => window.__hmr))).toBe(
-      bundlerSupportsHmrErrorRecovery,
-    );
+    expect(Boolean(await page.evaluate(() => window.__hmr))).toBe(bundlerSupportsHmrErrorRecovery);
   }),
 );
--- a/e2e/cases/hmr-yak-syntax-error-recovery/index.test.ts
+++ b/e2e/cases/hmr-yak-syntax-error-recovery/index.test.ts
@@ -3,8 +3,8 @@ import { withTestEnv } from "next-yak-e2e";
 
 test(
   "HMR recovers after a syntax error in a .yak.ts dependency",
-  withTestEnv("hmr-yak-syntax-error-recovery", async (fsTmp, page) => {
-    await page.goto(fsTmp.url);
+  withTestEnv("hmr-yak-syntax-error-recovery", async (testEnv, page) => {
+    await page.goto(testEnv.url);
 
     const box = page.getByTestId("box");
     await expect(box).toHaveCSS("padding", "40px");
@@ -16,7 +16,7 @@ test(
     });
 
     // Introduce a syntax error in the .yak.ts file
-    await fsTmp.writeFile(
+    await testEnv.writeFile(
       "tokens.yak.ts",
       "export const spacing = 5 * 8;\nexport const brand = <<<BROKEN>>>;\n",
     );
@@ -25,7 +25,7 @@ test(
     await page.waitForTimeout(3_000);
 
     // Fix the syntax error with new values
-    await fsTmp.writeFile(
+    await testEnv.writeFile(
       "tokens.yak.ts",
       'export const spacing = 2 * 8;\nexport const brand = "red";\n',
     );
@@ -34,7 +34,10 @@ test(
     await expect(box).toHaveCSS("padding", "16px", { timeout: 30_000 });
     await expect(box).toHaveCSS("color", "rgb(255, 0, 0)");
 
-    // Verify no full page reload occurred
-    expect(await page.evaluate(() => window.__hmr)).toBe(true);
+    /** rsbuild recovers from the error via a full page reload instead of a hot update */
+    const bundlerSupportsHmrErrorRecovery = testEnv.bundlerDirName !== "rsbuild";
+    expect(Boolean(await page.evaluate(() => window.__hmr))).toBe(
+      bundlerSupportsHmrErrorRecovery,
+    );
   }),
 );
--- a/packages/next-yak/rsbuild/index.ts
+++ b/packages/next-yak/rsbuild/index.ts
@@ -1,0 +1,92 @@
+import { fileURLToPath } from "node:url";
+import type { RsbuildPlugin } from "@rsbuild/core";
+import {
+  buildYakPluginOptions,
+  resolveYakContext,
+  type YakConfigOptions,
+} from "../withYak/index.js";
+
+// Rspack's loader context is webpack-compatible but does NOT implement
+// `this.loadModule` (the webpack API the webpack-loader relies on). It does
+// provide `this.fs.readFile`, `this.getResolve` and `this.addDependency` — which
+// is exactly what the turbopack loader uses — so the turbopack loader runs as-is
+// on Rspack. The compiled loader lives next to this file under dist/
+// (dist/rsbuild/index.js and dist/loaders/turbo-loader.cjs are siblings).
+const rspackLoaderPath = fileURLToPath(new URL("../loaders/turbo-loader.cjs", import.meta.url));
+
+/**
+ * Rsbuild plugin for next-yak.
+ *
+ * Wires the yak-swc transform as an Rspack `pre` loader that runs the
+ * compile-time CSS extraction in-loader and inlines the result as a
+ * `data:text/css;base64,` import (the same strategy as the Turbopack loader).
+ * Running it as a `pre` loader keeps yak-swc out of Rsbuild's own
+ * `builtin:swc-loader`, so it composes cleanly with that pass (TS/JSX, React
+ * Fast Refresh, and the React Compiler).
+ *
+ * @example
+ * ```ts
+ * // rsbuild.config.ts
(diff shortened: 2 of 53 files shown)


e2e/cases/hmr-yak-syntax-error-recovery/index.test.ts
```diff
@@ -34,7 +34,8 @@ test(
     await expect(box).toHaveCSS("padding", "16px", { timeout: 30_000 });
     await expect(box).toHaveCSS("color", "rgb(255, 0, 0)");
 
-    // Verify no full page reload occurred
-    expect(await page.evaluate(() => window.__hmr)).toBe(true);
+    /** rsbuild recovers from the error via a full page reload instead of a hot update */
+    const bundlerSupportsHmrErrorRecovery = testEnv.bundlerDirName !== "rsbuild";
```
Mad-Kat, line 38: Wasn't this something they wanted/did fix?

packages/next-yak/rsbuild/index.ts
```diff
@@ -0,0 +1,92 @@
...
+const rspackLoaderPath = fileURLToPath(new URL("../loaders/turbo-loader.cjs", import.meta.url));
+
+/**
+ * Rsbuild plugin for next-yak.
+ *
+ * Wires the yak-swc transform as an Rspack `pre` loader that runs the
+ * compile-time CSS extraction in-loader and inlines the result as a
+ * `data:text/css;base64,` import (the same strategy as the Turbopack loader).
+ * Running it as a `pre` loader keeps yak-swc out of Rsbuild's own
+ * `builtin:swc-loader`, so it composes cleanly with that pass (TS/JSX, React
```
Mad-Kat, line 24: can't we somehow add `yak-swc` as a plugin to the internal SWC transformation?

Mad-Kat (approved):

packages/next-yak/rsbuild/index.ts
jantimon: Good idea, that's actually how our `experimental.swcPlugins` Next.js integration works

I tried running yak-swc as a plugin inside Rspack's own builtin:swc-loader to get a single SWC pass like Next and added a small post-loader for our cross-file resolution

It works with all e2e green so I asked Claude to test the performance and it showed that the double SWC is surprisingly faster:

Isolated cold build (500 styled files): SWC single compile ~1.1s vs ~0.18s (~6x slower), wall time ~1.7x slower.
Repeated e2e (10 runs each): SWC single compile lost 2/10, avg 24.1s vs 21.8s 

Claude assumes that the gap is basically the Wasm runtime. yak-swc is a Wasm plugin in both modes, but our standalone loader runs it through @swc/core, which precompiles the Wasm to native via Wasmer and caches it (.swc/plugins/*.wasmer)... so execution is near-native. The same plugin directly in the rspack loader won't get that cached-native path so the per-file Wasm overhead dominates and eats the one parse advantage

maybe @kdy1 could verify wether that is correct or if Claude just made this up 😅

e2e/cases/hmr-yak-syntax-error-recovery/index.test.ts
jantimon: Yes in #508 @stormslowly tried to address this but I don't know the progress - I would go with the workaround for now.. if they fix it we can just update and remove the workaround

kdy1:
@jantimon Do you still my review? I was busy working on other proejcts so I was going to review this later, but it's merged so asking

jantimon:
@kdy1 no worry it is definitely not urgent - would be rather a funny learning for us and a potential follow up performance optimisation to run swc only once instead of twice

but the current solution works already quite nice

PR #565, opened 2026-06-30
Tighten import detection for compilation targets

Mad-Kat:
Gate the Turbopack loader and Vite plugin on a *real* `import "next-yak"` or `require "next-yak"` import instead of a loose `next-yak` substring match.

This was discovered while working with `fumadocs-mdx` on turbopack where  we compiled the output of it's compiler again and it resulted in broken code.

### DIFF
--- a/.changeset/spotty-cougars-hug.md
+++ b/.changeset/spotty-cougars-hug.md
@@ -1,0 +1,5 @@
+---
+"next-yak": minor
+---
+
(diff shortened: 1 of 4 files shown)


Mad-Kat:
As internally discussed: So far we don't see the necessity to do this step, but if users encounter issues in the wild we can speak about it again

PR #567, opened 2026-07-02
optimize css props into a pure static className string

jantimon:
The css prop compiles to a `__yak_mergeCssProp(...)` call that runs per element per render to compute a class string that is already known at build time. This folds it directly into the JSX:

```tsx
// before                                                                  // after
<div {...__yak_mergeCssProp({}, css("yX"))} />                             <div className="yX" />
<div {...__yak_mergeCssProp({}, css(() => on && css("yX1"), "yX"))} />     <div className={"yX" + (on ? " yX1" : "")} />
<div {...__yak_mergeCssProp({}, on ? css("yA") : css("yB"))} />            <div className={on ? "yA" : "yB"} />
```

Folding only happens when the result is provable. Elements with a `className` attribute, spread props, dynamic values (css variables) or mixin references keep the runtime merge exactly as before. `style` attributes stay in place (a class-only css prop never writes style), class order and condition evaluation order/count are identical to the runtime path, and the `/*YAK Extracted CSS:*/` comment stays anchored to the folded expression so CSS extraction keeps working in webpack, turbopack and vite.

In local measurements this makes server rendering of css-prop heavy pages roughly 1.3x to 1.7x faster (a 400 tile product listing went from ~93k to ~160k rendered instances per second, which is on par with hand written className strings). Pages that mostly use dynamic css values are unaffected.


packages/yak-swc/yak_swc/src/utils/css_prop.rs
```diff
@@ -40,7 +43,16 @@ impl CSSProp {
   ///     className: "myClassName"
   ///   })} />
   /// ```
-  pub fn transform(&self, opening_element: &mut JSXOpeningElement, merge_ident: &Ident) {
+  pub fn transform(&self, opening_element: &mut JSXOpeningElement, yak_imports: &mut YakImports) {
+    // Constant folding: when the css prop compiled to a statically known set of
+    // class names (and the element has no className/spread that would need
+    // runtime merging), replace the whole mergeCssProp call with a plain
+    // className attribute — the merge machinery is a per-element, per-render
+    // cost for a result that is known at build time.
```
Mad-Kat: do we need this comment? I would prefer to have the documentation on the function itself. That way you get information when you hover instead of while reading/looking through the code

packages/yak-swc/yak_swc/src/utils/css_prop.rs
```diff
@@ -73,6 +85,213 @@ impl CSSProp {
...
+  ///
+  /// ```jsx
+  /// <div css={css("a")} />                                // -> <div className="a" />
+  /// <div css={css(() => on && css("b"), "a")} />          // -> <div className={"a" + (on ? " b" : "")} />
+  /// <div css={on ? css("a") : css("b")} />                // -> <div className={on ? "a" : "b"} />
+  /// ```
+  ///
+  /// Class order matches the runtime processor exactly: the base class first,
+  /// then each conditional class in argument order. Conditions are still
+  /// evaluated once per render in the same order as before.
```
Mad-Kat:
```suggestion
  /// Tries to replace the css prop with a compile-time className attribute.
  ///
  /// Folds only when every part is statically known
  /// ```jsx
  /// <div css={css("a")} /> 
  /// <div css={css(() => on && css("b"), "a")} /> 
  /// <div css={on ? css("a") : css("b")} />
  /// ```
  ///
  /// Class order matches the runtime processor exactly. 
  /// Conditions are still evaluated once per render in the same order as before.
```

packages/yak-swc/yak_swc/src/utils/css_prop.rs
```diff
@@ -73,6 +85,213 @@ impl CSSProp {
...
+          JSXAttrName::Ident(ident) if ident.sym == *"style" => {}
+          _ => return false,
+        },
+        _ => return false,
+      }
+    }
+    let Ok(css_expr) =
+      Self::extract_css_expr(&opening_element.attrs[self.index], opening_element.span)
+    else {
+      // invalid css attribute: let the regular transform path report the error
```
Mad-Kat:
```suggestion
      // invalid css attribute
```

packages/yak-swc/yak_swc/src/utils/css_prop.rs
```diff
@@ -73,6 +85,213 @@ impl CSSProp {
...
+      return false;
+    };
+    let Some(folded) = Self::fold_css_expr(&css_expr, yak_imports) else {
+      return false;
+    };
+    match folded {
+      // an empty css`` contributes nothing — drop the attribute entirely
+      None => {
+        opening_element.attrs.remove(self.index);
+      }
```
Mad-Kat: How does this realistically happen?  I see the only case if a user writes

```jsx
<div css={css``} />
```

But should we really drop this? It was a deliberate choice by the user wasn't it?

Mad-Kat: Additionally it would make the code a bit easier to understand as `Option<Option<Box<Expr>>>` isn't the most readable way to represent this. If we really want to keep the functionality of dropping empty css classes we should use a more expressive type to distinguish the cases

packages/yak-swc/yak_swc/src/utils/css_prop.rs
```diff
@@ -73,6 +85,213 @@ impl CSSProp {
...
+
+  /// Folds a compiled css expression into a className expression.
+  /// Outer `None` = not statically foldable (keep the runtime path),
+  /// inner `None` = folds to nothing (empty css``).
+  ///
+  /// The folded expression reuses the span of the css call it replaces: the
+  /// `/*YAK Extracted CSS:*/` comment is attached to that span and is
+  /// LOAD-BEARING — `loaders/lib/extractCss.ts` parses it out of the
+  /// transformed source to build the stylesheet — so it must stay anchored to
+  /// an emitted node.
```
Mad-Kat:
```suggestion
  /// The folded expression reuses the span of the css call it replaces so the already present marker for CSS extraction also moves
```

packages/yak-swc/yak_swc/src/utils/css_prop.rs
```diff
@@ -73,6 +85,213 @@ impl CSSProp {
...
+      match arg.expr.unwrap_parens() {
+        Expr::Lit(Lit::Str(class_name)) => {
+          if base.is_some() {
+            return None;
+          }
+          base = Some(class_name.value.clone());
+        }
+        Expr::Arrow(arrow) => segments.push(Self::fold_condition_arrow(arrow, yak_imports)?),
+        // dynamic values (`{ style: … }`), mixin references, nested runtime
+        // logic — not foldable
```
Mad-Kat:
```suggestion
        // logic are not foldable
```

Mad-Kat (changes requested):
I'm not a fan of the heavy comments, but the logic looks ok with one case I would love to discuss: Should we remove empty `css` calls?

packages/yak-swc/yak_swc/src/utils/css_prop.rs
jantimon: Okay I removed it - good point 👍

jantimon:
> I'm not a fan of the heavy comments, but the logic looks ok with one case I would love to discuss: Should we remove empty `css` calls?

empty css calls are now handled in: #578

packages/yak-swc/yak_swc/tests/fixture/css-prop-style-reference/output.dev.stderr
Mad-Kat, line 1: do we want to mix error cases and valid cases in this test? Shouldn't we split it?

Mad-Kat (approved):

PR #568, opened 2026-07-03
fold static styled components into plain DOM elements

jantimon:
A fully static styled component still renders a wrapper component per element per render even though everything it does is known at build time. When it is declared and used in the same file, this folds the JSX usage directly into the DOM element:

```tsx
// before                              // after
const Card = styled.div`color: red;`;
<Card>hi</Card>                        <div className="yX">hi</div>
<Card className="user" />              <div className="yX user" />
<Card className={on && "on"} />        <div className={__yak_mergeClassNames("yX", on && "on")} />
```

Dynamic class-toggling styles fold too. The style conditions are inlined at the usage by substituting the prop values; a value the compiler cannot prove pure is bound once, so it still evaluates exactly as often as the original JSX did (the binding machinery is #595):

```tsx
const Box = styled.div<{ $on: boolean }>`
  ${({ $on }) => $on && css`color: red;`}
`;

// before                              // after
<Box $on />                            <div className="yB yA" />
<Box $on={open} />                     <div className={"yB" + (open ? " yA" : "")} />
<Box $on={isOpen(item)} />             <div className={(($on) => "yB" + ($on ? " yA" : ""))(isOpen(item))} />
```

(The first row is the after-minify form - the raw output is `"yB" + (true ? " yA" : "")`; constant folding is left to the minifier.)

A fully static `styled(Component)` folds to the wrapped component, which may live in another file:

```tsx
// before                              // after
const Fancy = styled(Card)`gap: 4px;`;
<Fancy>hi</Fancy>                      <Card className="yF">hi</Card>
```

A regular prop stays on the element, so when a condition reads one that is not provably pure, the element itself is wrapped and the value bound once - the className and the DOM attribute always read the same value:

```tsx
// before
<Button disabled={props.isBusy()}>Save</Button>

// after
{((disabled) => <button disabled={disabled}
    className={"yB" + (disabled ? " yA" : "")}>Save</button>)(props.isBusy())}
```

Folding only happens when the result is provable. Every "no" below keeps the runtime wrapper - a skipped usage costs performance, never correctness:

```mermaid
flowchart TD
  usage(["JSX usage of a styled component"]) --> sameFile{"declared in the same file as a<br/>top-level const Card = styled...?"}
  sameFile -- "no: import { Card } from './card'<br/>let Card = ..., or wrapped:<br/>const Card = memo(styled.div`...`)" --> keep["no fold - keep the runtime wrapper<br/>(slower, never incorrect)"]
  sameFile -- yes --> tagShape{"what is being styled?"}
  tagShape -- "styled.div.attrs(...)`...`<br/>styled('my-custom-element')`...`" --> keep
  tagShape -- "styled(Link)`...` containing<br/>${({ $active }) => $active && css`...`}<br/>($active is also forwarded to Link,<br/>which may style with it too)" --> keep
  tagShape -- "styled.div`...`, styled('div')`...`,<br/>fully static styled(Component)`...`" --> segments{"what do the ${...} pieces<br/>in the template do?"}
  segments -- "compute a css value:<br/>width: ${({ $w }) => $w}px<br/>(becomes a css variable in style)" --> keep
  segments -- "written as function (p) {...},<br/>or anything but toggling css" --> keep
  segments -- "nothing, or toggle css blocks:<br/>${({ $on }) => $on && css`...`}" --> attrs{"can the compiler see every prop<br/>passed at this usage?"}
  attrs -- "no: {...props} spread, generic type<br/>arguments, theme={t}, xlink:href,<br/>the same attribute written twice" --> keep
  attrs -- yes --> reads{"does every prop the conditions read<br/>come straight from a JSX attribute?"}
  reads -- "no: reads theme, children, className,<br/>style or key (injected or stripped by React),<br/>({ $on: x }) rename, = default, ...rest,<br/>or the whole props object escapes: calc(p)" --> keep
  reads -- yes --> purity{"can each attribute value run more<br/>than once without anyone noticing?"}
  purity -- "yes: $on={open}, $c={colors[status]},<br/>$odd={i % 2 !== 0}" --> inline["fold: condition inlined into className<br/>'yX' + (open ? ' yY' : '')"]
  purity -- "no, e.g. a call - but only<br/>className needs the value:<br/>$on={isOpen(item)}" --> iife["fold: value bound once in an IIFE<br/>(($on) => '...')(isOpen(item))<br/>erased by swc and terser"]
  purity -- "no, and the DOM needs it too:<br/>disabled={busy()}" --> wrap["fold: element wrapped, value bound<br/>once around it - className and DOM<br/>attribute read the same value"]
```

(One nuance the last question glosses over: a value that only *allocates* - an arrow, an object literal, JSX - still runs exactly once, it is just allowed to move, since creating a closure a little earlier is not observable. So `onClick={() => setOpen(true)}` never forces the element wrap.)

The declaration itself stays untouched for exports, selectors and non-foldable usages and is dead-code eliminated by the minifier when unused. A parity test verifies the folded output renders the exact same DOM as the wrapper component. One deliberate behavior change: foreign `$`-prefixed props on folded usages are forwarded to the DOM instead of being filtered by the runtime.

Stacked on #567

compile time fold code optimizations


docs/content/docs/faq.mdx
```diff
@@ -126,6 +126,81 @@ export default withYak(
...
+
+const App = () => <Icon $active={on} />;
+
+// compiles to
+const App = () => <span className={"Icon_a1b2" + (on ? " Icon_c3d4" : "")} />;
+```
+
+A `styled(Card)` usage is inlined to `Card` instead of a DOM element, so `Card` itself stays in the tree.
+
+One edge case is worth knowing about: if the same component is used both inlined and with a wrapper, switching between the two at runtime remounts the subtree instead of updating it, because React sees a different element type.
```
Mad-Kat, line 196: should this be a callout?

docs/playground-wasm/src/lib.rs
```diff
@@ -108,6 +108,7 @@ fn yak_pass(
                 .into(),
             false, // suppress_deprecation_warnings
             false, // react_refresh_reg
+            true,  // optimize_static_jsx
```
Mad-Kat, line 111: should this be a toggle in the playground like the 2 other options?

Mad-Kat (approved):

docs/content/docs/faq.mdx
jantimon: I believe this is a edge case where HMR state is lost (e.g. an unmanaged `<input>` loses its state) because you use for example a spread like `{...props}` or remove one

I don't think the information helps dev in any way that they try to avoid doing this to keep HMR in perfect shape

so rather no

docs/playground-wasm/src/lib.rs
jantimon: Agreed: follow-up issue #610 and follup pr in #612

PR #571, opened 2026-07-07
Add globalStyle for global styles

### LINKED ISSUE #376: mixin as a function result?
Hey, we have a following case.
We have an existing styled-components app which theme's has function for calculating spaces eg.
```
import styled from 'styled-components'

const Text = styled.div`
  ${({ theme }) => theme.getFluidSpacing('padding-left', 'equal', 0)}
`
```

What would be a replacement in next-yak?
I've tried following:
```
import {styled, css} from 'next-yak'
import { theme } from '@theme/main'
const spacing = css`${theme.getFluidSpacing('padding-left', 'equal', 0)}`

const Text = styled.div`
  ${spacing}
`
```

and it doesn't work. It shows the given css as classname, not as style.
```
<div class="index_Text__giUSk 
    padding-left: 4rem;

    @media screen and (min-width: 20em) {
      padding-left: calc(4rem + 0 * ((100vw - 176rem) / -144));;
    }

    @media screen and (min-width: 110em) {
      padding-left: 4rem;
    }
  ">Home Page</div>
```

How should it be done?

jantimon:
Adds a `globalStyle` tagged template for global styles (resets, `@font-face`, `:root` variables, view transitions). Closes the biggest gap when migrating from styled-components and replaces the deprecated `:global()` workaround with a real API. Closes #376.

```tsx
import { globalStyle, keyframes } from "next-yak";
import { Dialog } from "./dialog";

const fadeIn = keyframes`
  from { opacity: 0; }
`;

globalStyle`
  body {
    margin: 0;
  }

  ::view-transition-new(root) {
    animation: ${fadeIn} 200ms ease;
  }

  /* only yak can write this selector, the class name is compiler generated */
  body:has(${Dialog}[open]) {
    overflow: hidden;
  }
`;
```

The styles ride the same extraction pipeline as `styled` and `keyframes`: importing the file puts them in the stylesheet before first paint. There is no component to mount or unmount, conditions are expressed in CSS (`@media`, `@supports`, `:root[data-theme]`, `:has()`).

Design decisions:

- Import equals apply. The template compiles away, only a bare `globalStyle()` no-op call stays as anchor for the extracted CSS comment.
- Module scope only, enforced with a compile error (new `function_depth` tracking in the visitor).
- No runtime interpolations. `${props => ...}` is a compile error that points to the CSS custom property pattern instead. Constants, keyframes, static mixins and component selectors resolve at build time through the existing cross-file machinery.
- Output is emitted verbatim: unscoped, no `@layer` wrapping, no selector rewriting. Globals cascade against component styles by regular specificity, exactly like a hand written `globals.css`. An earlier draft wrapped everything in `@layer yak.global`, but layers sort before specificity, so a component's plain `color: black` would silently beat a state scoped global like `input:focus-visible { color: orange }`. Cascade layers stay available by authoring `@layer` inside the template, it passes through verbatim.
- Known limitation: bare class selectors for third party markup (`.maps { ... }`) work on all native CSS pipelines (Vite, Turbopack, rsbuild/rspack) but get scoped away by css-loader on Next.js webpack, where the non portable workaround is `:global(.maps)`. Accepted and documented rather than fixed with selector rewriting; separate file emission (#553) would dissolve it.

Testing: 8 fixture cases under `tests/fixture/global-styles` (including dynamic value and nesting errors), e2e case `e2e/cases/global-styles` green on all 7 bundlers in dev and prod. It covers the element selector, a `:root` variable consumed by a component, a keyframes reference, `body:has(<component>)`, the `:focus-visible` specificity win over a component style and an authored `@layer` losing to component styles.

Docs: new "Global styles" section in features.mdx, `createGlobalStyle` moved from Missing to Supported in the migration guide with a migration snippet.

Deferred follow ups: tree shaking hardening for global-only modules in packages with `sideEffects: false`, and an eslint rule for module scope usage.


.changeset/global-css.md
```diff
@@ -0,0 +1,6 @@
+---
+"yak-swc": minor
+"next-yak": minor
+---
+
+Add `globalCss` to write global styles that are extracted at build time and apply once the file is imported
```
Mad-Kat: Is "once the file is imported" really the users expected behaviour? It's a side effect and styled-components create a "normal" component out of it that can be placed, which is more explicit but also has it's own disadvantages (rendering two global styles, @import doesn't work...)

Mad-Kat: styled-components names it `createGlobalStyles`. Should we name it `createGlobalCss`?

docs/content/docs/features.mdx
```diff
@@ -341,6 +341,114 @@ animation: 1s var(--yakVar1) ease-out;
 </Tabs>
 </details>
 
+## Global styles
+
+For styles that don't belong to a single component — CSS resets, `@font-face` declarations, `:root` variables or view transitions — you can use the `globalCss` API. The styles are extracted at build time like all other yak styles and apply as soon as the file declaring them is imported, typically from your root layout or entry point.
```
Mad-Kat:
```suggestion
For styles that don't belong to a single component like CSS resets, `@font-face` declarations, `:root` variables or view transitions, you can use the `globalCss` API. The styles are extracted at build time like all other yak styles and applied as soon as the file declaring them is imported, typically from your root layout or page entry.
```

docs/content/docs/features.mdx
```diff
@@ -341,6 +341,114 @@ animation: 1s var(--yakVar1) ease-out;
...
+    animation: ${fadeIn} 200ms ease;
+  }
+`;
+```
+
+```jsx title="layout.tsx"
+import "./globals";
+```
+
+Global styles are part of the stylesheet, not part of the render tree. Unlike styled-components' `createGlobalStyle` there is no component to mount or unmount — express conditions in CSS instead (`@media`, `@supports`, `:root[data-theme]`, `:has()`).
```
Mad-Kat:
```suggestion
Global styles are part of the stylesheet, not part of the render tree. Unlike styled-components `createGlobalStyle` there is no component to mount or unmount.

If you need conditional styling, you can use CSS native functionality for that like `@media`, `@supports`, `:root[data-theme]` or `:has()`.
```

docs/content/docs/features.mdx
```diff
@@ -341,6 +341,114 @@ animation: 1s var(--yakVar1) ease-out;
...
+
+You can interpolate everything that is known at build time: constants, `keyframes`, static `css` mixins and even references to your styled components:
```
Mad-Kat:
```suggestion
You can interpolate everything that is known at build time like constants, `keyframes`, static `css` mixins and even references to your styled components:
```

docs/content/docs/features.mdx
```diff
@@ -341,6 +341,114 @@ animation: 1s var(--yakVar1) ease-out;
...
+
+globalCss`
+  /* prevent scrolling while the dialog is open */
+  body:has(${Dialog}[open]) {
+    overflow: hidden;
+  }
+`;
+```
+
+Runtime values (`${(props) => ...}`) are not supported inside `globalCss` — a global rule has no element to attach them to. Declare a CSS custom property instead and toggle it via an attribute or class on the root element:
```
Mad-Kat:
```suggestion
Runtime values (`${(props) => ...}`) are not supported inside `globalCss` as a global rule has no element to attach them to. Declare a CSS custom property instead and toggle it via an attribute or class on the root element:
```

docs/content/docs/features.mdx
```diff
@@ -341,6 +341,114 @@ animation: 1s var(--yakVar1) ease-out;
...
+```
+</Tab>
+</Tabs>
+</details>
+
+<Callout type="warn">
+  When using Next.js with webpack, class selectors for markup you don't control (e.g. a third-party
+  widget's `.maps` class) need to be wrapped in `:global()` to escape CSS Modules scoping:
+  `:global(.maps) {/* ... */}`. In native CSS mode (Vite, Turbopack, Rsbuild) the same selector must be
+  written without the wrapper — plain `.maps {/* ... */}` just works there.
```
Mad-Kat:
```suggestion
  The same rules for referencing external styles apply here as well. For further documentation visit: https://yak.js.org/docs/migration-from-styled-components#reference-external-styles
```

docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -354,6 +354,45 @@ export const mixin = css<{ $primary: boolean }>`
 `; // [!code --]
 ```
 
+### Global styles
+
+`createGlobalStyle` returns a component that you render — in next-yak global styles are a statement.
```
Mad-Kat:
```suggestion
`createGlobalStyle` returns a component that you render. In next-yak global styles are a statement.
```

docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -354,6 +354,45 @@ export const mixin = css<{ $primary: boolean }>`
...
+[`globalCss`](/docs/features#global-styles) applies as soon as the file declaring it is imported, so
+remove the JSX usage and import the file from your layout instead:
```
Mad-Kat: Don't we want to say that it should either be inlined where it's rendered (not declared) or imported at that position. It doesn't have to be the layout, but most often it will be

docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -354,6 +354,45 @@ export const mixin = css<{ $primary: boolean }>`
...
+        <GlobalStyle /> {/* [!code --] */}
+        {children}
+      </body>
+    </html>
+  );
+}
+```
+
+Theme props like `${(props) => props.theme.color}` inside global styles become CSS custom properties
+toggled via an attribute or class on the root element, and a conditionally rendered `<GlobalStyle />`
```
Mad-Kat:
```suggestion
toggled via an attribute or class on the root element. And a conditionally rendered `<GlobalStyle />`
```

e2e/cases/global-style/index.tsx
Mad-Kat, line 1: What would win if we specify a component style above global styles and target that again? Currently it's source order, but wouldn't it make sense to hoist global css? Isn't that the expected behaviour?

packages/next-yak/runtime/globalStyles.tsx
```diff
@@ -0,0 +1,62 @@
...
+ * Only build-time values are allowed — constants, `keyframes` animation names,
+ * static `css` mixins and styled-component selectors. Runtime functions
+ * (`${(props) => ...}`) are intentionally excluded: a global rule has no element
+ * to attach a CSS variable to. Declare a CSS custom property instead and toggle
+ * it via an attribute/class on the root element.
+ */
+export type GlobalCssInterpolation =
+  | string
+  | number
+  // keyframes`...` resolves to the animation name (a string) at build time
```
Mad-Kat: I would remove this comment

packages/yak-swc/yak_swc/tests/fixture/global-styles/error-dynamic-value/output.dev.tsx
```diff
@@ -0,0 +1,7 @@
+import { globalCss } from "next-yak/internal";
+import "./input.yak.module.css!=!./input?./input.yak.module.css";
+/*YAK Extracted CSS:
+body {
+  color: ;
+}
+*/ /*#__PURE__*/ globalCss();
```
Mad-Kat: should we revert the changes there were already made to the AST in this case? So it behave similar to the `only-global` test case

packages/yak-swc/yak_swc/tests/fixture/global-styles/only-global/output.prod.tsx
```diff
@@ -0,0 +1,9 @@
+import { globalCss } from "next-yak/internal";
+import "./input.yak.module.css!=!./input?./input.yak.module.css";
+// A module which only declares global styles and nothing else.
+// The side-effect CSS import must still be injected.
+/*YAK Extracted CSS:
+html {
+  color-scheme: light dark;
+}
+*/ /*#__PURE__*/ globalCss();
```
Mad-Kat: if this gets treeshaken, do we have the guarantee that the css import remains as it's a side-effect?

Mad-Kat: Can it be reordered so that we get specificity problems down the line?

packages/yak-swc/yak_swc/tests/fixture/global-styles/static/output.turbo.dev.tsx
```diff
@@ -0,0 +1,21 @@
...
+}
+:root {
+  --spacing: 4px;
+  --color-brand: #6b21ff;
+}
+body {
+  margin: 0;
+  font-family: sans-serif;
+}
+:global(.sr-only) {
```
Mad-Kat, line 17: shouldn't this be filtered out?

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -286,8 +298,9 @@ where
       css_state = Some(new_state);
 
       // Check for user-written :global() selectors in the raw quasi string
-      // This checks the original source code, not the transformed CSS
-      if !self.suppress_deprecation_warnings && !self.has_user_global {
+      // This checks the original source code, not the transformed CSS.
+      // Inside globalCss :global() is the documented escape hatch, not deprecated.
```
Mad-Kat: you sure? Why is `:global()` handled differently for global styles vs component styles? We compile to the same side effect import no?

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -499,6 +512,20 @@ ${{() => {var}}};\n",
...
+                  expr.span(),
+                  "Dynamic values are not supported in global styles because there is no element to attach them to.\n\
+                   Declare a CSS custom property instead and toggle it via an attribute/class on the root element:\n\n\
+                   globalCss`:root { --spacing: 4px; } :root[data-compact] { --spacing: 2px; }`\n\n\
+                   See https://yak.js.org/global-styles#dynamic",
+                )
+                .emit();
+            });
+            continue;
+          }
```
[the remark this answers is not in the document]
Mad-Kat: exactly what I saw in the snapshots

Mad-Kat (changes requested):
This PR is very solid and I found just some issues with it.

.changeset/global-css.md
jantimon: https://vanilla-extract.style/documentation/global-api/global-style/ `globalStyle`
https://mui.com/material-ui/experimental-api/pigment-css/ `globalCss`

styled-components + emotion + goober: `createGlobalStyles`

goober: glob: https://goober.js.org/api/createGlobalStyles/

jantimon: decided `globalStyles` for easier `createGlobalStyles` -> `globalStyles` migration

packages/yak-swc/yak_swc/tests/fixture/global-styles/only-global/output.prod.tsx
jantimon: yes it gets treeshaken and I was about to remove it entirely but we want to keep it as anchor for the css `/*YAK Extracted CSS:\n … */ ` comment

jantimon: reordering is a problem we can't fix - it is 1:1 the same as with styled.component order issues

packages/yak-swc/yak_swc/src/lib.rs
jantimon: yeah no real reason for the split so i unified it

globalStyles now emits the same `:global()` deprecation warning as styled/css, they all compile to the same side effect import and all steer people to native css transpilation mode

dropped the `inside_global_styles` guard and reframed the docs callout around native css mode

packages/yak-swc/yak_swc/tests/fixture/global-styles/static/output.turbo.dev.tsx
jantimon: this one is intentional and consistent with styled/css, we only unwrap `:global()` through webpack and native bundlers keep the selectors verbatim (reverted the native unwrap experiment in 807b66d for that reason)

on native pipelines you just write `.sr-only` directly

and with the unify change on the lib.rs thread `:global()` in globalStyles warns now too so its signposted instead of silently doing nothing

e2e/cases/global-style/index.tsx
jantimon: its the same source order model that already applies to component styles

both alternatives have bad tradeoffs, `@layer` sorts before specificity so a plain component `color` would beat a `:focus-visible` global (thats why we emit globals without a layer), and hoisting runs into the same unfixable ordering problem as styled component source order

so keeping source order as is for now

packages/next-yak/runtime/globalStyles.tsx
jantimon: done

docs/content/docs/migration-from-styled-components.mdx
jantimon: changed that please review

docs/content/docs/migration-from-styled-components.mdx
```diff
@@ -354,6 +354,47 @@ export const mixin = css<{ $primary: boolean }>`
 `; // [!code --]
 ```
 
+### Global styles
+
+`createGlobalStyle` returns a component that you render. In next-yak global styles are a statement.
+[`globalStyles`](/docs/features#global-styles) applies as soon as the file declaring it is imported, so
```
Mad-Kat: should we name it `globalStyle` as `createGlobalStyle` is the name styled-components use: https://styled-components.com/docs/api#createglobalstyle

sorry for the confusion

Mad-Kat (approved):
Looks good, the rename is up to you to decide. I thought the API of styled-components is `createGlobalStyles`, that's why I suggested `globalStyles` but as it's `createGlobalStyle`, `globalStyle` would be better and means less migration effort.

packages/yak-swc/yak_swc/src/yak_transforms.rs
```diff
@@ -529,6 +529,50 @@ impl YakTransform for TransformStyled {
...
 }
 
+/// Transform for global styles
+/// e.g. globalStyles`body { margin: 0; }`
+///
+/// The only transform without a name or scope: declarations are emitted
+/// verbatim — unscoped, unlayered and with untouched selectors in both
+/// transpilation modes. In `CssModule` mode this relies on css-loader leaving
+/// element/pseudo selectors alone; user class selectors that must stay global
+/// need an explicit `:global(.foo)` there.
```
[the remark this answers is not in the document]
jantimon: good catch, reworded the comment

the `:global()` you see there comes from interpolation resolution (the same path styled/css use), not from this transform which just emits the declarations verbatim

spelled that out now, user authored selectors stay as written while interpolated component/keyframes refs pick up the `:global(.hash)`/`global(name)` wrapper under css-loader

PR #572, opened 2026-07-08
Fix misplaced pure comment on inlined css mixins under wasm

jantimon:
Named `css` mixins inlined by identifier into a `styled` component could emit a stray `/*#__PURE__*/` at the end of the module. A downstream minifier (Rspack's SWC minifier) re-attaches that dangling comment to the preceding `export { ... }`, decides the export is pure + unused, and tree-shakes it away — so cross-chunk importers blow up with `TypeError: Cannot use 'in' operator ... in undefined`.

Root cause: `TransformNestedCss` gave its synthesized `css("hash")` call a `Span::dummy_with_cmt()` position and hung the pure annotation on that position via the comment side-table. That works natively (one process, shared comment map), but in the wasm plugin the node's byte-position and the comment's key travel through two separate ABI channels and stop lining up — the comment orphans and the host flushes it at the end of the file.

Fix is to stop keying pure on a fabricated position and carry it on the node's span instead (`PURE_SP`, same mechanism `assign_display_name` already uses). It survives cloning and serialization, needs no lookup, and codegen output is byte-identical to before. `lib.rs` guards the old comment-table fallback so pures don't pile onto the shared `BytePos::PURE` key. After this nothing in yak-swc calls `dummy_with_cmt()`.

Why the existing tests didn't catch it: `#[testing::fixture]` runs everything in-process, and the wasm-only code path isn't even compiled into the test binary. No native fixture can ever reproduce an ABI-boundary bug. So this also adds a wasm regression lane (`pnpm --filter yak-swc test:wasm-fixtures`) that runs the built `.wasm` through `@swc/core` over every fixture, asserts the pure invariants (no trailing/after-export stray, every `css(` keeps its pure), and diffs against the native snapshots. Wired into CI.

Verified: `cargo test` 296 pass, zero snapshot changes; wasm lane green.


jantimon:
Heads-up for later (not fixed here): the `css-prop-invalid` fixture behaves differently under a real `@swc/core` build than in the native tests.

`<div css="invalid" />` — native rewrites the invalid prop away and prints a warning. Through `@swc/core` the plugin leaves the original code untouched and the warning gets swallowed, so a user building with the wasm plugin sees neither the fix nor the error.

Out of scope for this PR (the wasm fixture lane just skips diagnostic fixtures), but worth a proper look on its own.

.changeset/quick-lions-attack.md
Mad-Kat, line 1: Should we split this in another PR?

packages/yak-swc/scripts/test-wasm-fixtures.mts
Mad-Kat, line 1: should we create a typescript file out of this?

Mad-Kat (approved):

PR #577, opened 2026-07-11
eslint-plugin-yak: sharper style-conditions messages

jantimon:
I asked codex to build a real next-yak app with our eslint setup and it couldn't fix the warnings properly. it hushed them by deleting the props and hand-rolling css variables via inline 'style', which is exactly what we don't want

The old message included slightly confusing parts e.g. "return a css literal" without examples

so this reworks the messages to show a concrete before/after built from the offending code, splits the confusing css case in two and cleans up two bits of wrong metadata

### enum ternary, number literals (the clock 'z-index')

```tsx
import { css, styled } from "next-yak";
const HandPivot = styled.span'
  z-index: ${({ $kind }) => $kind === "second" ? 4 : 3};
';
```

#### lint response:

```
'z-index' will be automatically compiled into a CSS variable because of the runtime condition, e.g. 'z-index: var(--h45cH)'.
For your fixed-set condition return a static css declaration like css'z-index: 4;' instead

next-yak compiles the static css declaration into a toggleable class name, resulting in better performance optimized JavaScript and HTML

Instead of:
  z-index: ${({ $kind }) => $kind === "second" ? 4 : 3}

write:
  z-index: 3;
  ${({ $kind }) => $kind === "second" && css'z-index: 4;'}
```

### enum ternary, string literals (quotes stripped)

```tsx
import { css, styled } from "next-yak";
const StartPill = styled.span'
  background: ${({ $starting }) => $starting ? "#d8b4fe" : "#f6c453"};
';
```

#### lint response:

```
'background' will be automatically compiled into a CSS variable because of the runtime condition, e.g. 'background: var(--h45cH)'.
For your fixed-set condition return a static css declaration like css'background: #d8b4fe;' instead

next-yak compiles the static css declaration into a toggleable class name, resulting in better performance optimized JavaScript and HTML

Instead of:
  background: ${({ $starting }) => $starting ? "#d8b4fe" : "#f6c453"}

write:
  background: #f6c453;
  ${({ $starting }) => $starting && css'background: #d8b4fe;'}
```

### boolean ternary

```tsx
import { css, styled } from "next-yak";
const Button = styled.button'
  color: ${({ $primary }) => $primary ? "white" : "black"};
';
```

#### lint response:

```
'color' will be automatically compiled into a CSS variable because of the runtime condition, e.g. 'color: var(--h45cH)'.
For your fixed-set condition return a static css declaration like css'color: white;' instead

next-yak compiles the static css declaration into a toggleable class name, resulting in better performance optimized JavaScript and HTML

Instead of:
  color: ${({ $primary }) => $primary ? "white" : "black"}

write:
  color: black;
  ${({ $primary }) => $primary && css'color: white;'}
```

### '&&' with a literal value (single line, no default branch)

```tsx
import { css, styled } from "next-yak";
const Box = styled.div'
  color: ${({ $on }) => $on && "red"};
';
```

#### lint response:

```
'color' will be automatically compiled into a CSS variable because of the runtime condition, e.g. 'color: var(--h45cH)'.
For your fixed-set condition return a static css declaration like css'color: red;' instead

next-yak compiles the static css declaration into a toggleable class name, resulting in better performance optimized JavaScript and HTML

Instead of:
  color: ${({ $on }) => $on && "red"}

write:
  ${({ $on }) => $on && css'color: red;'}
```

### '&&' with a non-literal value, falls back to the generic message

```tsx
import { css, styled } from "next-yak";
import { colors } from "./theme";
const Box = styled.div'
  color: ${({ $variant }) => $variant === "primary" && colors.primary};
';
```

#### lint response:

```
Arrow functions in next-yak styled/css literals should return either a css'...' literal (compiled to a CSS class) or a value derived from the component's props (a runtime CSS variable). Returning a constant does neither. If this is a fixed-set condition, move the declaration into the arrow function and return a css'...' literal. Keep the prop, don't hand-roll a CSS variable.
```

### '() => constant', generic message

```tsx
import { css, styled } from "next-yak";
const color = "red";
const Box = styled.div'
  color: ${() => color};
';
```

#### lint response:

```
Arrow functions in next-yak styled/css literals should return either a css'...' literal (compiled to a CSS class) or a value derived from the component's props (a runtime CSS variable). Returning a constant does neither. If this is a fixed-set condition, move the declaration into the arrow function and return a css'...' literal. Keep the prop, don't hand-roll a CSS variable.
```

### block-statement return of constants, generic message (fires per return)

```tsx
import { css, styled } from "next-yak";
const Box = styled.div'
  color: ${() => {
    if (variant === "primary") { return primary } else { return secondary }
  }};
';
```

#### lint response:

```
Arrow functions in next-yak styled/css literals should return either a css'...' literal (compiled to a CSS class) or a value derived from the component's props (a runtime CSS variable). Returning a constant does neither. If this is a fixed-set condition, move the declaration into the arrow function and return a css'...' literal. Keep the prop, don't hand-roll a CSS variable.
```

### css trap, case A: static value split from its property, move it in

```tsx
import { css, styled } from "next-yak";
const Chip = styled.span'
  color: ${({ $variant }) => $variant === "primary" && css'red'};
';
```

#### lint response:

```
'color' is a static property, but the arrow returns a css'red' literal (a toggleable style chunk, not a value) that next-yak can't splice into a property it already wrote. Move the whole declaration inside the css literal like css'color: red;' instead.

Instead of:
  color: ${({ $variant }) => $variant === "primary" && css'red'}

write:
  ${({ $variant }) => $variant === "primary" && css'color: red;'}
```

### css trap, case B: prop-derived value wrapped in css, drop the css

```tsx
import { css, styled } from "next-yak";
const Bar = styled.div'
  width: ${({ $w }) => css'${$w}px'};
';
```

#### lint response:

```
'width' wraps the runtime value '${$w}px' in a css'...' literal, which next-yak can't splice into a property it already wrote. Since the value comes from props, drop the css'' and return it directly so it compiles to a CSS variable.

Instead of:
  width: ${({ $w }) => css'${$w}px'}

write:
  width: ${({ $w }) => `${$w}px`}
```

### css trap, fallback: property already inside the css literal, generic message

```tsx
import { css, styled } from "next-yak";
const Chip = styled.span'
  color: ${({ $variant }) => $variant === "primary" && css'color: red'};
';
```

#### lint response:

```
The CSS property is outside the arrow function but its value returns a css'...' literal, which next-yak can't combine into a single declaration. Move the whole declaration inside the css literal so it holds the property too, or drop the css'' and return a plain runtime value.
```

### aliased css import is preserved in the suggestion

```tsx
import { css as cssYak, styled } from "next-yak";
const Box = styled.div'
  z-index: ${({ $kind }) => $kind === "second" ? 4 : 3};
';
```

#### lint response:

```
'z-index' will be automatically compiled into a CSS variable because of the runtime condition, e.g. 'z-index: var(--h45cH)'.
For your fixed-set condition return a static css declaration like cssYak'z-index: 4;' instead

next-yak compiles the static css declaration into a toggleable class name, resulting in better performance optimized JavaScript and HTML

Instead of:
  z-index: ${({ $kind }) => $kind === "second" ? 4 : 3}

write:
  z-index: 3;
  ${({ $kind }) => $kind === "second" && cssYak'z-index: 4;'}
```


Tyderion (approved):

docs/content/docs/eslint-plugin.mdx
```diff
@@ -52,6 +52,46 @@ const myDiv = styled.div`
...
+      "yak/css-nesting-operator": "error",
+      "yak/css-global-deprecated": "warn",
+      "yak/enforce-semicolon": "error",
+      "yak/style-conditions": "warn",
+    },
+  },
+]);
+```
+
+Requires `"type": "module"` in `package.json` and `jiti` installed to load the TypeScript config.
```
Mad-Kat: is this true?

Mad-Kat (approved):

docs/content/docs/eslint-plugin.mdx
jantimon: no only for old node.js version

PR #583, opened 2026-07-14
Fix multi-token delete on the string collector

jantimon:
The string-backed `ClassNames` collector in #558 couldn't delete multi-token strings: `delete("flex items-center gap-4")` split the stored value into tokens but compared each against the whole unsplit argument, so nothing matched and the classes stayed. `atoms()`-style code that adds a class group and later removes it silently kept both.

Splits the delete argument into a token set and filters against it. The `has()` guard stays, so it still only removes a group that's actually present as a contiguous run — matching the old `Set<string>` semantics where you can only delete what you added. Added unit tests for the multi-token, single-token, and partial-match cases.

Finding (3) from the perf-stack review.

### DIFF
--- a/packages/next-yak/runtime/__tests__/cssLiteral.test.tsx
+++ b/packages/next-yak/runtime/__tests__/cssLiteral.test.tsx
@@ -1,6 +1,6 @@
 // @ts-nocheck These are runtime tests and the external API isn't the runtime (after compile) API
 import type { YakTheme } from "../context";
-import { css } from "../cssLiteral";
+import { css, ClassNames } from "../cssLiteral";
 
 describe("cssLiteral css function", () => {
   describe("static CSS class names", () => {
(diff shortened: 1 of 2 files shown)


jantimon:
Closing this — the bug is real but unreachable, and the fix trades it for a worse one.

`delete()` has no callers in the runtime. `atoms` only adds, and it splits multi-token strings into single tokens first (`atoms.tsx:26,36`), so nothing in the codebase can produce the shape this targets. `ClassNameCollector` is not exported and `delete` is undocumented, so the only way in is user code reaching into the 2nd arg of an interpolation function.

More importantly it regresses against main. `main` seeds the Set with `className.split(" ")`, so every entry is a single token. Against a real Set:

| case | main (Set) | perf today | this PR |
| --- | --- | --- | --- |
| incoming className, delete a sub-sequence | `mx-auto flex items-center` | same | `mx-auto` |
| group added, then deleted as a group | `mx-auto` | no-op | same |
| token also present standalone | `flex` | no-op | `""` |

So it fixes row 2 but breaks row 1, which matches main today. The cause is that `has()` checks adjacency while `delete()` removes tokens from anywhere. Exact Set emulation is not possible from a string, since grouping is unknowable once the tokens are joined — that is inherent to the collector rewrite, not something this can fix.

The single-token `delete` on `perf` already matches main on every shape anything can actually produce, so leaving it alone is the better trade.

jantimon:
Correcting myself on one point above: I said `delete()` has no callers. That is wrong, and misleading in a way worth fixing for the record.

The collector is handed to *user* functions, and those are the intended callers. `atoms()` accepts a function that receives it, and there is a test doing exactly this — `cssProp.test.tsx:174` ("merge properties with nested css and atoms removing classes") calls `classNames.delete("input")`, and `styled.test.tsx:457` calls `classNames.has(...)`. So `has`/`delete` are supported API, not dead code.

That does not change the outcome. The reason to close is the semantics one: `has()` checks adjacency while a token-`Set` `delete` removes tokens from anywhere, which regressed a case that currently matches main, and exact `Set` fidelity is impossible once the tokens are joined into a string. The existing tests only delete single tokens, which this PR would not have changed either way.

PR #584, opened 2026-07-14
eslint-plugin-yak: allow class references outside declarations

jantimon:
Change for our linting

Before:

the style-conditions rule flagged local or imported css mixins returned through a variable e.g.:
${({ $isSelected }) => $isSelected && activeStyles}

got a false invalidRuntimeReturnValue error even though Yak compiles expressions outside CSS declarations as runtime classes.

After: the rule mirrors Yaks CSS parser state and only validates arrow return values inside CSS declarations, where values become CSS variables.

Local and imported class references now pass. Static references inside declarations like border-color or nested var() and calc() values are still reported.

Added coverage for logical and ternary class references, imports, pseudo selectors, comments, quotes, and nested CSS functions.

### DIFF
--- a/packages/eslint-plugin-yak/rules/styleConditions.ts
+++ b/packages/eslint-plugin-yak/rules/styleConditions.ts
@@ -13,7 +13,7 @@ export const styleConditions = createRule({
     type: "suggestion",
     docs: {
       description:
-        "Enforces that arrow functions only return runtime values or css literals in styled/css literals from next-yak",
+        "Warns when arrow functions in next-yak styled/css literals would create unnecessary or invalid CSS variables",
       recommended: true,
       requiresTypeChecking: false,
     },
@@ -49,10 +49,7 @@ export const styleConditions = createRule({
           return;
         }
 
-        const codeBefore = getQuasiBeforeExpression(tag, needle);
-        // Guess that if a quasi ends with a colon that it is a declaration e.g.
-        // css`color: ${({ color }) => color}`
-        if (codeBefore === undefined || !codeBefore.trim().endsWith(":")) {
+        if (!isInsideCssDeclarationValue(tag, needle)) {
           return;
         }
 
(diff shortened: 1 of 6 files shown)


packages/eslint-plugin-yak/rules/styleConditions.ts
```diff
@@ -93,6 +90,13 @@ export const styleConditions = createRule({
           return;
         }
 
+        // Outside a declaration value Yak treats the expression as a runtime class
+        // expression, so returning local or imported css mixins is valid and cannot
+        // accidentally create a CSS variable.
+        if (!isInsideCssDeclarationValue(tag, needle)) {
+          return;
+        }
```
Mad-Kat, line 98: Why do we have to check this twice? Isn't it sufficient here?

Mad-Kat (approved):

packages/eslint-plugin-yak/rules/styleConditions.ts
jantimon: `styleConditions.ts:52` gates the css-literal visitor (invalidCssReturnValue). It only fires when the arrow returns a nested css literal:

```tsx
// outside a declaration → line 52 bails (valid toggleable class)
styled.button`
  ${({ variant }) => variant === "primary" && css`color: red`}
`;

// inside a declaration → line 52 passes and it reports invalidCssReturnValue
styled.div`
  margin: var(--foo, ${() => css`4px`});
`;
```

styleConditions.ts:96 gates the arrow/return visitor (invalidRuntimeReturnValue). It covers arrows that contain no css tag at all, so the first visitor never sees them:

```tsx
// outside a declaration → line 96 bails (mixin reference becomes a class)
styled.button`
  ${({ $active }) => $active && activeStyles}
`;
```

styleConditions.ts:109 is only reached when the line 96 gate passes — inside a declaration the return value becomes a runtime CSS variable and gets validated:

```tsx
// inside a declaration → gate passes, line 109 validates the return value
styled.button`
  color: ${({ $active }) => ($active ? "red" : "black")};
`;
```

So removing the check at 52 would flag every valid cond && css…`` mixin, and removing it at 96 would bring back the false positive on activeStyles references

PR #588, opened 2026-07-14
Don't substitute key into folded style expressions

jantimon:
React strips `key` before the component sees props, so `${(p) => p.key === activeId && css`…`}` reads `undefined` on the runtime path — but the fold substituted the JSX attribute value, so `<Item key="active" />` folded to `"active" === "active"` and applied styling the runtime path never would. Same source, different rendering depending on whether the usage folded, and foldability flips on unrelated edits.

Adding `key` to `is_runtime_injected_prop` reuses the existing bail-out: a style expression reading `key` just keeps the runtime wrapper, where React's stripping applies. Passing `key={…}` at a call site still folds — only *reading* it bails, so the cost is zero for real code. Fixtures cover both directions.

Finding (16) from the perf-stack review.


Mad-Kat (approved):
Who would possible write such code 😅

PR #589, opened 2026-07-14
Document $-prop eval counts in the folding changeset

jantimon:
Folding inlines a `$`-prop into every style expression that reads it, so `<B $v={compute()} />` runs `compute()` once per expression — and a `$`-prop no expression reads is dropped without running at all, so `<Toggle $on $log={track()} />` never calls `track()`. Both are accepted (see finding 1), but nothing said so.

Adds a sentence to the changeset's behavioral notes. Scoped it to `$`-props on dynamic usages after checking the code: the drop is gated on `!runtime_expressions.is_empty()` + `is_transient_prop`, so the broader "prop expressions ... may run zero times" would have contradicted the preceding sentence about static usages forwarding foreign `$props`. Also says "pure" rather than "side-effect free" — `is_duplication_safe` requires a stable value too, and `$v={Math.random()}` is side-effect free but still breaks.

Finding (1) from the perf-stack review.


jantimon:
Superseded: the $-prop eval-count explanation is going into the FAQ (on top of #590's rewrite) instead of the changeset. It's too much detail for a changelog line, and the FAQ has room to say which half StrictMode actually covers.

Finding (1) stays addressed, just in docs rather than the changeset.

PR #595, opened 2026-07-14
Evaluate every prop expression exactly once

jantimon:
next-yak renders every styled component as a small wrapper component around the DOM element it styles. #568 adds a build time optimization (`optimizeStaticJsx`): when a styled component is declared and used in the same file, the usage is folded into the plain DOM element and the wrapper never runs:

```tsx
const Sticker = styled.div<{ $tilt: number }>`
  ${(p) => p.$tilt > 5 && css`transform: rotate(3deg);`}
  ${(p) => p.$tilt > 8 && css`box-shadow: 0 0 8px black;`}
`;

<Sticker $tilt={7} />
// ships as
<div className="a1 a2" />
```

To fold a usage with a dynamic prop, the compiler substitutes the prop value into every style condition that reads it. During review of #568 it came up that this substitution changes how often the prop expression runs. #568 shipped that as a documented tradeoff, with an FAQ entry telling users to precompute such values.

This PR (stacked on #568) removes the tradeoff. A folded usage now evaluates every prop expression exactly once, in attribute order, same as the unfolded JSX. The FAQ entry is deleted because there is nothing left to warn about.

## What goes wrong today

JSX attribute position is a guarantee: `<Sticker $tilt={Math.random() * 10} />` runs the expression exactly once. Substituting the value into the conditions breaks that. Both conditions read `$tilt`, so both get a copy:

```tsx
<Sticker $tilt={Math.random() * 10} />

// folded output before this PR: two random numbers
<div className={"a1"
  + (Math.random() * 10 > 5 ? " a2" : "")
  + (Math.random() * 10 > 8 ? " a3" : "")} />
```

The sticker can get the shadow without the rotation, even though a tilt above 8 is always above 5. The wrapper component could never render that.

Duplication is not the only case. A condition can also read the prop inside a callback:

```tsx
const Row = styled.div<{ $n: number }>`
  ${({ $n }) => list.some((x) => x > $n) && css`color: red;`}
`;

// Math.random() runs once per render (twice in dev strict mode)
<Row $n={Math.random()} />
```

With a naive folding approach the value is copied into the condition:

```tsx
// Math.random() now runs once per list.some() iteration
<div className={"r1" + (list.some((x) => x > Math.random()) ? " r2" : "")} />
```

Or the read can sit behind a short circuit, and the value vanishes entirely:

```tsx
const Note = styled.div<{ $b: boolean }>`
  ${({ $b }) => flag && $b && css`color: blue;`}
`;

// Math.random() runs once per render
<Note $b={Math.random()} />

// naive fold: Math.random() never runs while flag is false
<div className={"n1" + (flag && Math.random() ? " n2" : "")} />
```

Note that both cases read the prop only once: an expression does not have to be duplicated for its evaluation count to change, so the fix has to cover every read, not just repeated ones.

## Why not just skip these usages

Skipping the optimization whenever a prop value is not provably pure would be safe and simple, but it gives up the reason folding exists. In the SSR benchmark the folded build renders 2.6x faster than next-yak 9.5 (193k vs 73k renders/s, 234 vs 130 req/s under sustained load) and removes the styling runtime from the profile entirely, leaving react-dom plus user code. Every skipped usage falls back toward the old numbers. So the goal is to fold these values correctly, not to stop folding them.

## The fix: bind the value instead of copying it

During substitution every prop read becomes a parameter. Values that are free to inline (literals, identifiers, member reads, comparisons of those) are put back where they are read, which reproduces today's output byte for byte. Whatever stays a parameter is bound once at the usage site, in argument position, which is the original guarantee.

The Sticker through the pipeline:

```tsx
// user input
<Sticker $tilt={Math.random() * 10} />

// yak output
<div className={(($tilt) =>
  "a1" + ($tilt > 5 ? " a2" : "") + ($tilt > 8 ? " a3" : "")
)(Math.random() * 10)} />

// after the swc minifier (Next.js production, @swc/core 1.15)
let $tilt;
<div className={"a1"
  + (($tilt = 10 * Math.random()) > 5 ? " a2" : "")
  + ($tilt > 8 ? " a3" : "")} />
```

The minifier dissolves the IIFE into a temp variable, so the Next.js path ships zero extra closures. terser does the same. React Compiler outlines the arrow to a module scope `_temp($tilt)` and memoizes the usage as usual (checked on the real output). esbuild keeps the closure, one allocation per render.

Static usages are untouched: `<Sticker $tilt={7} />` has nothing to bind and still compiles to `<div className="a1 a2" />`. Even a literal that does end up in an IIFE is constant-folded through it by swc, but esbuild will not fold through an IIFE. That is why pure values keep inlining instead of "bind everything and let the minifier clean up": on Vite that would trade the static single string literal for a closure.

## When the value has to reach the element too

A regular prop stays on the element, so binding it inside the className is not enough. The element itself gets wrapped:

```tsx
// user input
<Button disabled={props.isBusy()}>Save</Button>

// yak output
{((disabled) =>
  <button disabled={disabled} className={"b1" + (disabled ? " b2" : "")}>
    Save
  </button>)(props.isBusy())}

// after the swc minifier
let disabled;
disabled = props.isBusy(),
<button disabled={disabled} className={"b1" + (disabled ? " b2" : "")}>Save</button>
```

This is the button from the old FAQ that could render disabled while styled as enabled. The className and the DOM attribute now always read the same value.

The wrap only happens because `props.isBusy()` is a call, which the compiler cannot prove pure. With a plain boolean prop the value is safe to read twice, so nothing gets wrapped:

```tsx
// user input
<Button disabled={props.isBusy}>Save</Button>

// yak output
<button disabled={props.isBusy}
  className={"b1" + (props.isBusy ? " b2" : "")}>Save</button>
```

## Three output forms, picked so the common cases pay nothing

| how common | prop value | output form |
| --- | --- | --- |
| dominant | `$tilt={tilt}`, `$c={colors[status]}`, `$a={i % 4 !== 0}` | inline, byte-identical to #568 |
| common | `$active={isSelected(item)}` | className IIFE, erased by swc and terser |
| occasional | `disabled={f()}` on a dynamic usage | element wrap |
| rare | an effectful value ordered between two bound ones | element wrap |

The purity judgment (`utils/purity.rs`) is an allowlist with two levels, and anything it does not recognize counts as impure:

- **may run 0..N times**: literals, identifiers, member reads, and unary/binary combinations of those. These inline.
- **may move, but runs exactly once**: arrows, array/object literals, JSX. Creating a closure is not observable, so `onClick={() => setOpen(true)}` never forces the element wrap. Without this level nearly every interactive element would get wrapped for nothing.
- **impure**, including without any call involved: `$i={i++}`, `$d={new Date()}`, `$x={(y = f())}`.

## What still skips the optimization

A skipped usage keeps the wrapper component, so it costs performance but never correctness. Unchanged from #568: declared in another file or not a top-level `const`, conditions reading `theme`/`children`/`className`/`style`/`key`, `.attrs()`, css variables, `styled(Component)`, spreads, a `theme` prop, type arguments, destructuring with rename/default/rest.

Two new skips in this PR, both found during review: a namespaced attribute (`xlink:href`) and a repeated attribute. The fold moves all bound values to a single evaluation point and has to prove nothing observable sits between that point and where each value originally was. These two shapes hide an attribute from that analysis, so they now skip the optimization instead of silently reordering. `xlink:href` is reachable in practice: `<use xlink:href>` is the SVG sprite pattern and `svg`/`use`/`image` are foldable elements. Style conditions written as `function` expressions (they carry their own `this`/`arguments`) already skip it since #598.

Two escapes remain, both limited to code that React's own purity rule already forbids: a `$prop` that no condition reads is dropped without being evaluated, and a getter behind a member read (`$a={obj.x}`) can still fire once per condition. Gating member reads would cost `colors[status]` its inlining, which is not worth it.

## The eslint rule that appears and disappears in this branch

The commit history of this branch builds an eslint rule `precompute-style-prop-values` and later deletes it again. It never existed on main and the final diff does not touch eslint-plugin-yak at all.

The rule warned when a `$prop` expression would be inlined more than once. It was the wrong tool three times over: counting reads cannot see the callback and short-circuit cases above, its allowlist classified `i % 4 !== 0` as unsafe (the most common dynamic prop shape there is), and React already has [a purity rule](https://react.dev/reference/rules/components-and-hooks-must-be-pure#components-and-hooks-must-be-idempotent) that flags the actual problem: `$tilt={Math.random() * 10}` is an impure render with or without yak. Now that the compiler binds instead of copies, there is no yak-specific evaluation rule for users to know. The fold is also transparent to React's rule: values never move across it and parameters run exactly once, so the folded output violates the rule if and only if the source did.

React Compiler adoption does not make the binding redundant, though. React-pure is weaker than duplication-safe (local mutation and fresh allocations are allowed during render), and the callback case is a perf bug even for a perfectly pure value: `$n={expensivePure(x)}` read inside `list.some(...)` goes from O(1) to O(n).

## Implementation and review notes

Two source files plus fixtures:

- `utils/purity.rs` (new): the two-level purity judgment, one AST walk per value.
- `utils/styled_jsx_fold.rs`: the substitution and the three output forms. `select_shape` (attributes and bindings in, one of the three forms out) is pure data to data, so the whole case analysis is unit tested directly instead of through fixtures.
- The `styled-jsx-dynamic-folding` fixture grows all the new cases, the FAQ entry goes away, the changeset gains the exactly-once guarantee.

Design notes for the diff:

- Parameters are handed out during substitution instead of counting reads in a separate pass, so "was this prop read" is exact by construction. A separate counting pass would have been one more copy of the substitution logic, and the existing copies had already drifted apart.
- The element wrap captures every value the usage cannot move, not only the props the conditions read. That keeps all arguments in attribute source order: `<Button id={g()} disabled={f()} />` still runs `g` before `f`.
- An element cannot wrap itself, so the wrap is emitted by whoever owns the JSX node. Three visitor hooks cover the positions an element appears in: expression, JSX child, and unbraced attribute value.

Verified by 364 Rust tests (disabling the purity check fails all four fixture snapshot modes) plus a byte-for-byte diff of all 252 fixtures through the real `.wasm` build, which is green for the first time. Going green required dropping the `<Card<any>>` fixture case: TSX type arguments do not survive the wasm ABI, so the case pinned a native-vs-wasm difference with no behavior behind it.


packages/yak-swc/yak_swc/src/utils/purity.rs
```diff
@@ -0,0 +1,416 @@
...
+//! attribute value of a usage: [`purity`] answers them in a single walk and
+//! the caller stores the level - a value is never traversed twice to ask the
+//! other question.
+//!
+//! The levels are an allowlist: an unrecognised expression is
+//! [`Purity::Impure`]. The asymmetry is deliberate - a wrong answer towards
+//! pure is silent, a wrong answer towards impure only costs an IIFE parameter.
+//!
+//! Every value is judged under [`unwrap_type_casts`], so `f() as number` is
+//! judged on `f()` and `(x)` on `x`.
```
Mad-Kat: I prefer having comments on the code that does things, but I'm not inherently against it on top of the file. I just think it doesn't create much value

packages/yak-swc/yak_swc/src/utils/styled_jsx_fold.rs
```diff
@@ -327,12 +411,49 @@ fn is_transient_prop(attr: &JSXAttrOrSpread) -> bool {
...
+  /// every consumed value could be put back where it is read: today's output
+  Inline,
+  /// the values which stay bound are all `$`-props, which only the className
+  /// has to see, and nothing observable sits between them
+  ClassNameIife,
+  /// a bound value also has to reach the element, or the parameter block would
+  /// have to jump an observable evaluation: the element itself is wrapped, and
+  /// captures everything in source order
+  ElementWrap,
+}
```
Mad-Kat, line 460: Isn't this exactly the enum value meanings of `Purity`? Should we map these or from those?

Mad-Kat (approved):

packages/yak-swc/yak_swc/src/utils/purity.rs
jantimon: 👍 will do a comment rework on the prop branch

PR #600, opened 2026-07-16
Shrink the minified mergeClassNames

jantimon:
A minimal optimization to reduce the minified size of `mergeClassNames` (−12 B min / −22 B brotli). Behavior is unchanged, runtime perf is at parity — `mergeClassNames` only runs when class merging is actually needed, not on every render.


Mad-Kat (approved):
⛳

PR #601, opened 2026-07-16
Shrink the yak component symbol assignment

jantimon:
Attach the `yakComponentSymbol` tuple with a direct property write instead of `Object.assign` (−22 B min / −22 B brotli). Behavior is unchanged — the symbol is an own enumerable property either way, so `sym in Component`, the chain-merge destructure and selector targeting all still work. It also runs ~11× faster at module-load (no intermediate object), where every styled-component definition hits it once; the render path is untouched.

Type safety is kept via an internal `YakComponentInternals` tuple type, so the write is type-checked with no `as unknown` and the public `YakComponent` contract still exposes the opaque `[unknown, unknown, unknown, unknown]`.

### DIFF
--- a/packages/next-yak/runtime/styled.tsx
+++ b/packages/next-yak/runtime/styled.tsx
@@ -46,6 +46,16 @@ const styledFactory: StyledFn = (Component) =>
  */
 export const styled = styledFactory as Styled;
 
+// Real shape of the yakComponentSymbol tuple. Public YakComponent keeps it
+// opaque ([unknown, ...]); this internal type lets the write below be
+// type-checked without an `as unknown` cast.
+type YakComponentInternals = [
+  self: React.FunctionComponent,
+  attrsFn: AttrsFunction<any, any, any> | undefined,
+  styleProcessor: RuntimeStyleProcessor<unknown>,
+  target: React.FunctionComponent | string,
+];
+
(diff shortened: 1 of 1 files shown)


packages/next-yak/runtime/styled.tsx
```diff
@@ -46,6 +46,16 @@ const styledFactory: StyledFn = (Component) =>
  */
 export const styled = styledFactory as Styled;
 
+// Real shape of the yakComponentSymbol tuple. Public YakComponent keeps it
+// opaque ([unknown, ...]); this internal type lets the write below be
+// type-checked without an `as unknown` cast.
```
Mad-Kat:
```suggestion

```

Mad-Kat (approved):

packages/next-yak/runtime/styled.tsx
```diff
@@ -46,6 +46,15 @@ const styledFactory: StyledFn = (Component) =>
  */
 export const styled = styledFactory as Styled;
 
+// Real shape of the yakComponentSymbol tuple. Public YakComponent keeps it
+// opaque ([unknown, ...]); this internal type lets the write below be
```
jantimon:
```suggestion
// Real internal shape of the yakComponentSymbol tuple. Public YakComponent keeps it
// opaque ([unknown, ...])
```

PR #602, opened 2026-07-16
Shorten internal render markers via a namespace module

jantimon:
The three render-time marker keys (`$__attrs`, `$__runtimeStylesProcessed`, `$staticAttrs`) are pure internal state on props / the attrs fn — not a compiler contract (nothing in yak-swc emits them), never public. This lifts them into `internals/propMarkers.ts` as short `as const` values and uses them via `import * as INTERNAL`.

esbuild inlines the namespace away, so the app bundle ships only the short literals: **−70 B min / −20 B brotli** vs `perf`. Source reads better (the const name is the doc) and the marker string lives in one place instead of being hand-duplicated across 11 sites.

`$` stays load-bearing (`removeNonDomProperties` strips it); the `__` on the two prop markers keeps them from colliding with a user `$`-prop.

Verified locally: `tsc` clean, 188 runtime tests pass, built bundle contains only the short literals (namespace fully inlined).

### DIFF
--- a/packages/next-yak/runtime/internals/propMarkers.ts
+++ b/packages/next-yak/runtime/internals/propMarkers.ts
@@ -1,18 +1,4 @@
-/**
- * Internal, render-time markers written onto props (or the merged attrs
- * function) and read back to skip repeated work across a `styled(...)` chain.
- *
- * These are NOT a compiler contract — nothing in yak-swc emits them — and they
- * never appear on the public API. The leading `$` is load-bearing:
- * `removeNonDomProperties` strips `$`-prefixed props, so a marker can never
- * reach the DOM. The `__` on the two prop markers guards against a user's own
- * `$`-prop colliding with one. The values are kept short so the app bundler
- * ships fewer bytes; the descriptive export names are how they read in source.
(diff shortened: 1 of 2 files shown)


packages/next-yak/runtime/internals/propMarkers.ts
```diff
@@ -0,0 +1,34 @@
...
+ * `removeNonDomProperties` strips `$`-prefixed props, so a marker can never
+ * reach the DOM. The `__` on the two prop markers guards against a user's own
+ * `$`-prop colliding with one. The values are kept short so the app bundler
+ * ships fewer bytes; the descriptive export names are how they read in source.
+ *
+ * Imported as a namespace (`import * as INTERNAL`) so each use site reads as
+ * `INTERNAL.ATTRS_MERGED`; esbuild inlines the `const` values, leaving nothing
+ * of the namespace in the shipped bundle.
+ */
+
```
jantimon:
```suggestion
/** Internal, render-time markers written onto props */

```

Mad-Kat (approved):

PR #604, opened 2026-07-16
Bump @rsbuild/core to 2.1.6 for parallel SWC Wasm plugin compilation

jantimon:
Rspack 2.1.4 compiles SWC Wasm plugins in parallel with Wasmtime (~5x faster cold load): https://x.com/jiahan_c/status/2077771822976868585


jantimon:
we will do a dependency update pr soon

PR #607, opened 2026-07-22
Gate both folds behind foldStatic and run e2e in both modes

jantimon:
Once the folds land, e2e only sees optimized output — the runtime path loses coverage. This renames `optimizeStaticJsx` to `foldStatic` (unreleased, so no break), gates the css-prop fold under the same flag, and runs the full e2e matrix twice: a 2-job CI matrix sets `YAK_E2E_FOLD_STATIC`, read only by the e2e bundler configs — the published packages get no env hook.

Also fixes `viteYak` dropping `foldStatic` when resolving options, which made the off switch a no-op for vite.

SWC fixtures now run the runtime path by default; a fixture opts into folding with a `config.json` (`{"foldStatic": true}`). Two tricky fixtures pin the fold-vs-bail boundaries for styled JSX and css props, and a unit test pins the shipped default (on).


jantimon:
## Test plan
- [ ] CI: both matrix jobs green — `E2E tests (foldStatic: on)` and `(foldStatic: off)`
- [ ] Off job logs show `fold mode: off` and the `foldables` case passes
- [ ] `YAK_E2E_FOLD_STATIC=false pnpm --filter next-yak-e2e test vite` passes locally
- [ ] `cargo test` in packages/yak-swc/yak_swc green (fold-off fixtures)
- [ ] vite app with `viteYak({ foldStatic: false })` emits runtime wrappers, not folded divs

.changeset/cssprop-constant-folding.md
```diff
@@ -2,4 +2,4 @@
 "yak-swc": minor
 ---
 
-Fold statically known css props into a plain `className` at build time instead of emitting a per-render `__yak_mergeCssProp` call, e.g. `<div css={css`color: red`} />` becomes `<div className="yX" />` and conditional styles become `className={"yX" + (on ? " yX1" : "")}`. Elements with an existing `className`, spread props, dynamic values or mixin references keep the runtime merge path unchanged.
+Fold statically known css props into a plain `className` at build time instead of emitting a per-render `__yak_mergeCssProp` call, e.g. `<div css={css`color: red`} />` becomes `<div className="yX" />` and conditional styles become `className={"yX" + (on ? " yX1" : "")}`. Elements with an existing `className`, spread props, dynamic values or mixin references keep the runtime merge path unchanged. Set `foldStatic: false` to route every `css` prop that produces styles through the runtime path.
```
Mad-Kat:
```suggestion
Fold statically known css props into a plain `className` at build time instead of emitting a per-render `__yak_mergeCssProp` call, e.g. `<div css={css`color: red`} />` becomes `<div className="yX" />` and conditional styles become `className={"yX" + (on ? " yX1" : "")}`. Elements with an existing `className`, spread props, dynamic values or mixin references keep the runtime merge path unchanged.
```

packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -1501,7 +1532,7 @@ mod tests {
         import_mode,
         false,
         false,
-        true,
+        options.fold_static,
```
Mad-Kat, line 1546: should we do something like this also for the other configuration options? That way the snapshot test configuration isn't split in two places as it would be after this PR

Mad-Kat (approved):

packages/yak-swc/yak_swc/src/lib.rs
jantimon: nice idea 👍 did that

PR #608, opened 2026-07-22
Split e2e CI into parallel dev and build jobs

jantimon:
With two fold modes, each CI job runs dev then build tests in sequence (~12 min). A mode × phase matrix runs 4 jobs in parallel instead; wall time drops to the slowest job (dev, ~7 min). Setup steps stay identical per job — no artifact sharing.

### DIFF
--- a/.github/workflows/e2e.yml
+++ b/.github/workflows/e2e.yml
@@ -28,7 +28,7 @@ concurrency:
 
 jobs:
   e2e:
-    name: "E2E tests (foldStatic: ${{ matrix.fold-static }})"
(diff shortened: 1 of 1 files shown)


jantimon:
## Test plan
- [ ] All 4 matrix jobs green: (dev|build) × (foldStatic on|off)
- [ ] Dev jobs skip the build-test step and vice versa (check step logs)
- [ ] Wall time of the workflow ≈ slowest job, not the sum

Mad-Kat (approved):

PR #613, opened 2026-07-23
Collapse fully static styled() chains at the declaration

jantimon:
A fully static `styled(Card)` chain kept one runtime wrapper per authored level even when everything is known at build time. This collapses the declaration itself — `Fancy = styled(Card)`…`` becomes the base element with parent-first merged classes — so same-file usages fold to plain elements and every consumer of an exported chain gets a single wrapper; unused base components become dead code the minifier drops.

```tsx
const Button = styled.button`
  padding: 8px;
`;
export const FullWidthButton = styled(Button)`
  width: 100%;
`;
const App = () => <FullWidthButton>Save</FullWidthButton>;

// compiles to
export const FullWidthButton = /*#__PURE__*/ __yak.__yak_button("yak-button yak-full");
const App = () => <button className="yak-button yak-full">Save</button>;
```

The folding code is restructured into `utils/fold/` (registry + register/resolve/rewrite phases, usage rewrite, shared css-to-className primitive) so the whole feature reads as one design; the collapse is a step in the resolve phase. A skipped collapse is never incorrect: dynamic, `.attrs`, cross-file, mutable, and out-of-declaration-order parents keep today's behavior.

Compile-time cost: the collapse adds no full AST traversal. Chain resolution is registry-only hashmap work; `immutable_bindings` and `collapse_declarations` scan only top-level `module.body` items, and only a collapsed declarator's init expression gets visited (a few nodes, self-disarming visitor). The one full-depth pass (`usage::fold`) predates this PR and its body moved verbatim. With `foldStatic: false` or no registered components, `fold_module` returns before touching the AST. Added memory is a small `Vec<Collapse>` plus the interned merged class strings.


jantimon:
## Test plan
- [ ] `cargo test` green (390) — chain fixtures show collapsed declarations in all 4 modes
- [ ] `pnpm --filter yak-swc test:wasm-fixtures` 272/272 (PURE + extracted-CSS comments survive the wasm ABI)
- [ ] `foldedJsxParity` chain cases: folded output ≡ runtime output
- [ ] e2e `foldables` 3-level chain passes with foldStatic on and off
- [ ] Bail fixtures: OfDynamic/OfAttrs/OfLet/OfLater/ExtendedImport keep pre-collapse output

PR #614, opened 2026-07-23
Decode className strings merged by the css prop

jantimon:
A `css` prop next to a string `className` merges at runtime through `__yak_mergeCssProp`. The transform cloned the className into the emitted JS with its JSX raw, and JSX raws keep HTML entities encoded and backslashes literal — so real-world class names corrupted silently or broke the build:

```tsx
<div css={css`color: blue;`} className="Food &amp; Drink" />
// before: class="Food &amp; Drink"   (entity reaches the DOM encoded)
// after:  class="Food & Drink"

<div css={css`color: blue;`} className="before:content-['\2713']" />
// before: build failure — every bundler's loader re-parses the emitted JS and
//         rejects the raw \27 as a strict-mode octal escape
// after:  class="before:content-['\2713']"
```

The merge now builds the string from the decoded value (`raw: None`), the same construction every other Str site in the crate uses and the styled fold adopted in #605. The octal-escape build breaker is pinned by the SWC fixture in all four modes; e2e asserts the DOM class and applied style, using a non-digit backslash token because cold JSX dependency scanners (rolldown) reject `\<digit>` in raw source — a bundler limitation unrelated to yak.

### DIFF
--- a/.changeset/cssprop-classname-decoding.md
+++ b/.changeset/cssprop-classname-decoding.md
@@ -1,0 +1,5 @@
+---
+"yak-swc": patch
+---
+
+A `className` string next to a `css` prop kept its JSX encoding: `className="Food &amp; Drink"` rendered the entity into the DOM class, and a backslash class like `before:content-['\2713']` broke the build with a SyntaxError. The value now reaches the DOM exactly as written.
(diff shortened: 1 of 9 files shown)


jantimon:
## Test plan
- [ ] `cargo test` green — `css-prop` fixture shows `Food & Drink` decoded and `\\2713` double-escaped in emitted JS, all 4 modes
- [ ] `pnpm --filter yak-swc test:wasm-fixtures` green
- [ ] e2e `css-prop` case: entity class and backslash class reach the DOM as written, css-prop color applies to both

Mad-Kat (approved):
:shipit:

PR #615, opened 2026-07-23
Emit folded className merges in expression position

jantimon:
The fold merged a user `className` string into a JSX attribute string. swc prints attribute strings with JS escaping, but the JSX-preserve pipelines (Vite, Turbopack, rsbuild) re-parse the printed output — and JSX attribute strings take backslashes literally. A backslash class doubled and silently lost its style:

```tsx
const Card = styled.div`color: red;`;
<Card className="before:content-['\00d7']" />
// before: <div className="yX before:content-['\\00d7']" />  (re-parsed class has two backslashes)
// after:  <div className={"yX before:content-['\\00d7']"} /> (JS re-parse yields one backslash)
```

The merged value now goes through expression position, so print and re-parse share one escaping regime — the same direction the #605 fix established for attribute strings moved into expressions. The remaining attribute-string emission (generated class names only) is guarded by a `debug_assert` that the name carries no backslash or quote.

### DIFF
--- a/packages/yak-swc/yak_swc/src/utils/fold/usage.rs
+++ b/packages/yak-swc/yak_swc/src/utils/fold/usage.rs
@@ -484,8 +484,10 @@ impl FoldVisitor<'_> {
         // backslash or quote: those print differently under JS and JSX. A
         // merge routes through expression position for the same reason
         YakClassName::Static(class_name) => {
+          // scan the raw WTF-8 bytes: `\` and `"` are ASCII, so this never
+          // decodes the value and leaves any emoji or surrogate untouched
           debug_assert!(
-            !class_name.to_string_lossy().contains(['\\', '"']),
+            !class_name.as_bytes().iter().any(|&b| b == b'\\' || b == b'"'),
             "a folded class name must carry no backslash or quote: {class_name:?}"
           );
           JSXAttrValue::Str(str_lit(class_name, DUMMY_SP))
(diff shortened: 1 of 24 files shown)


jantimon:
## Test plan
- [ ] `cargo test` green (390) — string-attrs fixture pins the backslash merge as `className={"..."}` in all 4 modes, debug_assert active
- [ ] `pnpm --filter yak-swc test:wasm-fixtures` 272/272
- [ ] e2e `foldables`: folded backslash merge renders a single literal backslash, matches the runtime spread twin, with foldStatic on and off

packages/yak-swc/yak_swc/src/utils/fold/usage.rs
```diff
@@ -480,7 +480,16 @@ impl FoldVisitor<'_> {
     // leave the specifier behind with nothing referencing it
     let Some(index) = class_name_index else {
       let value = match yak_class {
-        YakClassName::Static(class_name) => JSXAttrValue::Str(str_lit(class_name, DUMMY_SP)),
+        // safe as a raw attribute string only while the class name carries no
+        // backslash or quote: those print differently under JS and JSX. A
+        // merge routes through expression position for the same reason
+        YakClassName::Static(class_name) => {
+          debug_assert!(
+            !class_name.to_string_lossy().contains(['\\', '"']),
```
Mad-Kat: classnames could potentially be emojis right? That would make it show � according to https://docs.rs/swc_atoms/latest/swc_atoms/struct.Wtf8Atom.html#method.to_string_lossy

Mad-Kat (changes requested):

packages/yak-swc/yak_swc/src/utils/fold/usage.rs
jantimon: Good catch to check — emoji are safe: `to_string_lossy` only substitutes U+FFFD for unpaired surrogates, and emoji are valid UTF-8. A surrogate can only reach the merge through a JS string literal (`className={"\uD800"}`) — a JSX attribute string keeps the escape as six literal characters. Still, 8844bb65 removes the lossy calls entirely: `merge_class_names` now concatenates at the WTF-8 level (`Wtf8Buf`), so any user class name is copied byte for byte. Pinned by a fixture (`🔥 mark` in all 4 modes), unit tests incl. the surrogate case, and an e2e twin asserting the emoji reaches the DOM.

Mad-Kat (approved):

PR #619, opened 2026-07-23
Add a folding guide to the docs

jantimon:
Folding is hard to get into from the option docs alone. This adds a guide that teaches it the way PR #568's description did: before/after pairs for each fold, a decision-tree diagram where every "no" keeps the runtime wrapper, and the caveats a user can observe (DevTools, `child.type`, `$`-props, evaluation guarantees). The diagram is a build-time mermaid render — the .mmd source sits next to the committed SVG with its regen command, no runtime dependency.

how-does-it-work and the FAQ shrink to summaries linking here. Also corrects the css-prop changeset: a mixin reference is a compile error, not a runtime-merge case.


jantimon:
## Test plan
- [ ] Docs build green; /docs/folding renders with the decision SVG
- [ ] Every before/after pair matches current compiler output (checked against fixtures in review)
- [ ] how-does-it-work and FAQ link to the page, no duplicated deep-dives

Mad-Kat (approved):

PR #626, opened 2026-08-17
Deduplicate identical function interpolations

### LINKED ISSUE #625: Identical interpolations create duplicate CSS variables
I noticed that when the same arrow function shows up twice in one styled component, next-yak creates a separate CSS variable for each one - even when they are exactly the same.

I ran into this with a scroll driven carousel where the animation range needs the same props twice:

## Reproduction case

```tsx

const Dot = styled.div<{ $index: number; $slideCount: number }>`
  animation-range:
    max(0%,   calc((${({ $index }) => $index} - 1) * 100% / (${({ $slideCount }) => $slideCount} - 1)))
    min(100%, calc((${({ $index }) => $index} + 1) * 100% / (${({ $slideCount }) => $slideCount} - 1)));
`;
```

## Actual behavior

The compiled output sets four CSS variables, although there are only two different functions:

```js
"style": {
    "--Dot__animation-range_m7uBBu":    ({ $index }) => $index,
    "--Dot__animation-range_m7uBBu-01": ({ $slideCount }) => $slideCount,
    "--Dot__animation-range_m7uBBu-02": ({ $index }) => $index,
    "--Dot__animation-range_m7uBBu-03": ({ $slideCount }) => $slideCount
}
```

Each of those runs on every render and every one ends up as its own entry in the style attribute on the DOM element.

## Expected behavior

Two variables would be enough:

```js
"style": {
    "--Dot__animation-range_m7uBBu":    ({ $index }) => $index,
    "--Dot__animation-range_m7uBBu-01": ({ $slideCount }) => $slideCount
}
```

The same happens across declarations too, e.g. `left: ${({ $x }) => $x}px; right: ${({ $x }) => $x}px;` gives two variables where one would do.

Would be great if the compiler could reuse the variable when two interpolations are exactly the same.

jantimon:
Fixes #625

Two interpolations of one styled component now share a CSS variable when they are the same at runtime. The compiler compares a normalized copy of each function: TypeScript casts, type annotations and parentheses are erased, and destructured parameters are sorted by key where reordering is provably safe (no default values, no computed keys). The emitted code always keeps the first occurrence untouched. Trailing units stay part of the identity, so `${fn}px` and `${fn}rem` keep separate variables.

The lookup is a plain linear scan over the interpolations of one component. Hashing doesn't pay off here: a component has a handful of interpolations, and the derived hashes of the swc AST include spans and raw text, so a hash map would disagree with the span-ignoring comparison. swc's own `NodeIgnoringSpan` makes the same call and falls back to linear comparison.

One note on variable counting: a reused interpolation allocates no name, so the `-01`, `-02` counters of later variables in the same file can shift. No existing fixture changed - only files with duplicate interpolations are affected.

The new fixture also surfaced a small pre-existing bug: a parenthesized arrow like `${(({ $z }) => $z)}` reported "must be wrapped in arrow functions" although it compiled fine. The validator now unwraps type casts and parentheses first.


packages/yak-swc/yak_swc/src/lib.rs
```diff
@@ -692,10 +681,41 @@ ${{() => {var}}};\n",
...
+              .and_then(|normalized| css_variable_dedup.find(normalized))
+              .map(str::to_string);
+
+            let css_variable_name = reused_name.unwrap_or_else(|| {
+              let readable_name = format!(
+                "{}__{}",
+                // Current variable name of the StyledComponent or Mixin
+                // e.g. Button for const Button = styled.button`color: red;`
+                self
+                  // TODO: check if parts should also be used for the name:
```
Mad-Kat, line 702: Should we keep the todo?

Mad-Kat (commented):
Only one comment. Otherwise looks small and like a good optimisation

Mad-Kat (approved):

packages/yak-swc/yak_swc/src/lib.rs
jantimon: Good catch - the TODO is older than this PR, the diff only moved it. It asks whether the variable name should use the full component path like the class name already does: `const buttons = { primary: styled.button` gets the class `buttons_primary` but the variable `--buttons__color`. Resolved in #629 - the variable now uses the same path.

PR #631, opened 2026-08-18
feat(eslint-plugin-yak): support Oxlint JS plugins

usercao:
## Summary

Add direct Oxlint JS plugin support to `eslint-plugin-yak` without introducing a separate `oxlint-plugin-yak` package.

## Motivation

The current package is coupled to the ESLint ecosystem even when it is only loaded by Oxlint:

- Its runtime code imports `@typescript-eslint/utils`.
- It declares `eslint`, `@typescript-eslint/parser`, and `typescript` as required peer dependencies.

As a result, an Oxlint-only consumer must install the ESLint stack in order to install and load `eslint-plugin-yak`. This PR removes that requirement for Oxlint usage while preserving the existing ESLint integration.

- Keep `eslint-plugin-yak` as the single published package and preserve ESLint support through `eslintCompatPlugin`.
- Move the rules to Oxlint's `createOnce` API so Oxlint can optimize rule initialization and traversal.
- Reset per-file `next-yak` import state in `before()` to prevent aliases from leaking between files.
- Replace the production dependency on `@typescript-eslint/utils` with `@oxlint/plugins` and make `eslint` an optional peer dependency.
- Add Oxlint configuration documentation, unit coverage using Oxlint's `RuleTester`, and a CLI integration fixture that loads the package through `jsPlugins`.
- Add a minor changeset for `eslint-plugin-yak`.

## Oxlint usage

```jsonc
// .oxlintrc.json
{
  "jsPlugins": ["eslint-plugin-yak"],
  "rules": {
    "yak/css-nesting-operator": "error",
    "yak/css-global-deprecated": "warn",
    "yak/enforce-semicolon": "error",
    "yak/style-conditions": "warn"
  }
}
```

## Compatibility notes

`eslintCompatPlugin` adds ESLint-compatible `create` methods for the new `createOnce` rules, so the existing ESLint API and recommended config continue to work.

Oxlint-only consumers no longer need to install ESLint. ESLint consumers still install ESLint as usual. `next-yak` is only a workspace development dependency used to resolve the CLI test fixture's realistic `import { styled } from "next-yak"` source.

## Validation

- `pnpm --filter eslint-plugin-yak exec tsc --noEmit`
- `pnpm --filter eslint-plugin-yak test -- --watch=false` - 6 test files, 122 tests passing
- `pnpm --filter eslint-plugin-yak check:docs`
- `oxfmt --check` and `git diff --check`
- Packed `eslint-plugin-yak` and installed it into an isolated consumer with only `oxlint` plus the tarball. `eslint` was absent from `node_modules`, while Oxlint loaded `eslint-plugin-yak` from `jsPlugins` and reported `yak(css-nesting-operator)` as expected.


Mad-Kat:
Hey @usercao 

Thank you so much for creating this. This is definitely the direction we want to go as oxlint becomes more and more standard. 

We will have a look :)

packages/eslint-plugin-yak/package.json
```diff
@@ -47,26 +48,33 @@
   },
   "scripts": {
     "build": "tsdown",
+    "pretest": "tsdown",
```
Mad-Kat:
```suggestion
    "pretest": "pnpm build",
```
This way we can freely change the build command and don't need to keep it in sync

packages/eslint-plugin-yak/package.json
```diff
@@ -47,26 +48,33 @@
...
-    "@typescript-eslint/utils": "catalog:dev"
+    "@oxlint/plugins": "catalog:dev"
   },
   "devDependencies": {
     "@types/node": "catalog:dev",
     "@typescript-eslint/parser": "catalog:dev",
     "@typescript-eslint/rule-tester": "catalog:dev",
+    "eslint": "catalog:dev",
     "eslint-doc-generator": "catalog:dev",
+    "next-yak": "workspace:*",
```
Mad-Kat: This isn't needed and can safely be removed

packages/eslint-plugin-yak/rules/oxlintCli.test.ts
```diff
@@ -0,0 +1,47 @@
...
+const oxlintBin = join(dirname(require.resolve("oxlint/package.json")), "bin/oxlint");
+const fixtureDirectory = fileURLToPath(new URL("./fixtures/oxlint-cli/", import.meta.url));
+
+describe("Oxlint CLI integration", () => {
+  it("loads the package and resets import state between files", () => {
+    const result = spawnSync(
+      oxlintBin,
+      [
+        "--config",
+        ".oxlintrc.json",
```
Mad-Kat: Currently the `lint` step in CI fails because it tries to lint the new fixture. If you rename the config in the fixture to `config.json` so it's not autodiscovered by oxlint and add `packages/eslint-plugin-yak/rules/fixtures/**` to the to root `.oxlintrc.json` `ignorePatterns`, it should work.

packages/eslint-plugin-yak/rules/oxlintCli.test.ts
```diff
@@ -0,0 +1,47 @@
...
+        "--format",
+        "json",
+        "input.ts",
+        "without-import.ts",
```
Mad-Kat:
```suggestion
        ".",
```
This way we don't need to keep the files in the fixture and here as arguments in sync. It could also help us if we ever want to add more cli tests.

Mad-Kat (changes requested):
Thank you for this excellent PR. Just some nitpicks and one issue with linting. Otherwise it's good to go :)

usercao:
Hi @Mad-Kat, thank you for the review. I've addressed all of your comments in `02ca58c`. Could you please take another look and let me know if there is anything else I should adjust?

Mad-Kat (approved):

Mad-Kat:
@usercao Thank you very much for this PR and your effort. If you find anything else or just want to dabble with the codebase you can always open a PR. We're always happy about help :)

Mad-Kat:
Should be released now in version `1.3.0`. Thanks again!

PR #632, opened 2026-08-19
Add @yak/solid for solid v2

Mad-Kat:
This PR adds a new public package `@yak/solid`. The first package that uses the new org. The runtime is solid v2 only and the SWC plugin can now handle `classname` and `class` differences as well as basic folding but that is disabled as default for now.

In order to share the core (vite loader logic, cross-file & node evaluation worker) I've created a new internal only package `packages/yak-internals` that is bundled in the `next-yak` and `@yak/solid` packages. 

The differences to the react version are documented in the readme and are:
- `useTheme()` returns an accessor: read `theme().highContrast`, not `theme.highContrast`. The same applies to `props.theme()` in interpolations and `.attrs()`
- `class` instead of `className`, also inside `.attrs({ ... })`
- No jsx-runtime export: Solid compiles JSX natively. The css prop types activate by importing `@yak/solid`
- HMR handling is very similiar but we add the pragma `/* @refresh component */` to signalize a component boundary for exported components to Solid's compiler


Mad-Kat:
The e2e test currently fails because the HMR story for Solid is a bit different than for React.

There are 2 issues currently. 

**First, why the test fails:** the compiled `const Divider = styled("hr"...)` doesn't look like a component to Solid, so its compiler doesn't inject the HMR boundary. We had the same issue for React, but in #524 we added `$RefreshReg$` markers for exactly this case. We'd have to implement something similar for Solid ([reference is their new Rust compiler based on oxc](https://github.com/ryansolid/dom-expressions/tree/next/packages/compiler/__tests__/refresh/fixtures)).

**The second issue:** if a user only updates styles (and not any JS), the current module invalidation would still invalidate the JS module, and Solid would update at the HMR boundary. This is the same for React, but since components can keep their state in React during HMR, it's not a big issue (just a very small performance penalty). In Solid, however, it's a bigger issue, since Solid can't keep the signal state after HMR and resets the component. I think this experience can be jarring, and we should fix it. One solution would be to have a small cache in the Vite plugin that tracks the CSS module transforms and invalidates just that CSS module instead of the whole JS module. But I'm not a big fan of adding that logic to the plugin, as it also has major footguns if we don't invalidate correctly.

Mad-Kat:
Turns out we couldn't do the same for Solid in HMR and self register our call signatures because it's compiler does all the work, it doesn't have a runtime part like in React and didn't have a mechanism for us to extend it. That's the reason I created https://github.com/solidjs/solid/issues/3090 and Ryan fixed it very fast. I've updated to `rc4`, changed the plugin to emit `/* @refresh component */` and HMR works now 🎉

This also fixed the need for the CSS cache in the vite plugin. I will create another PR to improve the HMR story & performance in vite where this can be reintroduced, but I want this PR to focus on necessary changes for solid.

Mad-Kat:
Another big step forward. Ryan solved also my other opened issue: https://github.com/solidjs/solid/issues/3099 which solves the issue with `foldStatic`. I've updated our usage of Solid to `rc5`, enabled folding again and added the two PR's to our yak shaving section.

packages/yak-internals/loaders/lib/parseExports.ts
```diff
@@ -0,0 +1,206 @@
+import { parse } from "@babel/parser";
+import type { Expression, ObjectExpression, TSAsExpression } from "@babel/types";
+import {
+  ModuleExport,
+  ModuleExports,
+  UnsupportedExportSource,
+} from "../../cross-file-resolver/parseModule.js";
+
+export async function parseExports(sourceContents: string): Promise<ModuleExports> {
```
jantimon, line 17: why do we need this what does this function do?

exported function should always have a /** */ comment

packages/yak-internals/loaders/lib/parseExports.ts
```diff
@@ -0,0 +1,206 @@
...
+  try {
+    const ast = parse(sourceContents, {
```
jantimon, line 19: should we skip parsing this file if `sourceContents.includes("yak") === false`?

.changeset/yak-package-autodetect.md
```diff
@@ -0,0 +1,6 @@
+---
+"yak-swc": minor
+"next-yak": minor
+---
+
+Support the new `@yak/solid` package: the compiler now detects which yak package a file imports from. Custom bundler integrations can import the cross-file resolver from `next-yak/cross-file-resolver` and the `.yak` file evaluator from `next-yak/isolated-source-eval`.
```
jantimon:
```suggestion
Add support for `SolidJs` together with the new `@yak/solid` package
```

.changeset/yak-solid-initial.md
```diff
@@ -0,0 +1,5 @@
+---
+"@yak/solid": minor
+---
+
+Initial release of `@yak/solid`, the SolidJS runtime for yak, targeting Solid 2 (RC). Provides `styled`, `css`, `keyframes`, `globalStyle`, `atoms`, the css prop and a signal-based theme (`useTheme` returns an accessor) with build-time CSS extraction via the yak SWC plugin and `yak` from `@yak/solid/vite`.
```
jantimon:
```suggestion
Initial release of `@yak/solid` the SolidJS runtime for yak targeting Solid 2 with full build time css extraction
```

jantimon (approved):

packages/yak-internals/loaders/lib/parseExports.ts
Mad-Kat: Yes :) I think the added comment makes it clearer. 

This is a 1:1 move from our `packages/next-yak/loaders/lib/resolveCrossFileSelectors.ts` but I had to split it up because it mixed core logic with a webpack specific function

packages/yak-internals/loaders/lib/parseExports.ts
Mad-Kat: No we can't as a `export const primary = "purple"` still needs to be put in a file that uses `next-yak`. So we can't skip these files for cross file handling

PR #637, opened 2026-08-31
Handle css props strict where possible

### LINKED ISSUE #635: css prop accepts invalid values: array crashes at runtime, object renders unstyled
### Describe the bug

With `strictCssProp` at its default (`true`), both of these compile instead of failing:

```tsx
import { css } from "next-yak";

const pad = css`
  padding: 8px;
`;

// 1. compiles, then throws at render:
export const App = () => <button css={[pad]}>click</button>;

// 2. compiles, renders with no styles and no warning:
export const Alt = () => <div css={{ color: "red" }}>hello</div>;
```

Case 1 throws `TypeError: cssProp is not a function` from `mergeCssProp`, which receives the array and calls it as a function.

Case 2 renders this, with no error at any stage:

```html
<div css="[object Object]">hello</div>
```

### Expected behavior

Both should fail the build under `strictCssProp`. The TypeScript types already reject them (`CSSProp` allows only a single `ComponentStyles` plus `false`/`null`/`undefined`), so tsc with `jsxImportSource: "next-yak"` is currently the only guard, and eslint-plugin-yak misses both too.

### Environment

- next-yak / yak-swc: 9.8.0
- @rsbuild/core 2.1.13 via `pluginYak()`; the transform is shared across bundlers, so other loaders are likely affected too (not separately verified)
- React 19.2, Node 24

Mad-Kat:
Closes: #635

When `strictCssProp` (default: `true`) is enabled, we check the values that are passed now more thoroughly. The SWC plugin checks if only a valid function e.g. `css` or `atoms` or a logical combination thereof is used.

For cases where static analysis can't reach (e.g. spreading) we throw an error during runtime in DEV if we encounter any wrongly passed values.

### DIFF
--- a/packages/yak-swc/yak_swc/src/utils/fold/css_prop.rs
+++ b/packages/yak-swc/yak_swc/src/utils/fold/css_prop.rs
@@ -74,60 +74,64 @@ impl CSSProp {
     if fold_static && self.try_fold(opening_element, yak_imports) {
       return;
     }
-    let merge_ident = yak_imports.get_yak_utility_ident("mergeCssProp");
-    // Build the replacement from borrowed attributes first, without mutating the
-    // element. That way an invalid `css` prop can be left completely untouched
-    // when strict mode is off (it might belong to another library, not next-yak).
-    let merge_call: Result<Box<Expr>, TransformError> = (|| {
-      let css_expr =
-        Self::extract_css_expr(&opening_element.attrs[self.index], opening_element.span)?;
-      Self::validate_css_value(&css_expr, yak_imports)?;
-      let mapped_props = self
-        .relevant_props
-        .iter()
-        .map(|&(index, _)| match &opening_element.attrs[index] {
-          JSXAttrOrSpread::JSXAttr(attr) => Self::map_jsx_attr(attr),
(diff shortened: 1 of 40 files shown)


packages/yak-swc/yak_swc/src/utils/fold/css_prop.rs
```diff
@@ -81,9 +81,7 @@ impl CSSProp {
     let merge_call: Result<Box<Expr>, TransformError> = (|| {
       let css_expr =
         Self::extract_css_expr(&opening_element.attrs[self.index], opening_element.span)?;
-      if let Some(span) = Self::find_style_reference(&css_expr) {
-        return Err(TransformError::StyleReference(span));
-      }
+      Self::validate_css_value(&css_expr, yak_imports)?;
```
jantimon: is this strict mode only?

jantimon (approved):

packages/yak-swc/yak_swc/src/utils/fold/css_prop.rs
Mad-Kat: Good question. It ran in both cases but it reported it only in `strict` mode. I've updated the code and renamed the error to make this distinction clearer

PR #639, opened 2026-09-02
Fix JSX parsing in turbopack

### LINKED ISSUE #638: Turbopack: .jsx files with next-yak styled fail to parse
Hi,

First of all, thanks for maintaining next-yak and for the great work on the project. I've been trying it with the latest Next.js/Turbopack setup and came across an issue with .jsx files.

Environment
-----------

*   Next.js: `16.3.3`
*   Next-Yak: `9.8.0`
*   React: `19.2.8`
*   Node.js: `26.8.1`
*   Pnpm: `10.11.0`

Reproduction
------------

The issue occurs when a `.jsx` file contains a next-yak `styled` component:

    const Container = styled.div`
      display: flex;
    `;
    
    export default function Component() {
      return <Container>Hello</Container>;
    }
    

With Turbopack, this fails with an SWC error such as:

    Expression expected
    

The same code works correctly with Webpack and when using `.tsx`.

A `.jsx` file without a next-yak `styled` component does not reproduce the issue.

The issue affects both `next dev` and `next build`.

Investigation
-------------

The issue appears to come from `createTransform()` in:

`packages/next-yak/loaders/turbo-loader.ts`

The original `@swc/core.transform()` configuration does not specify `jsc.parser`.

Adding the following locally fixes the issue:

    parser: {
      syntax: "ecmascript",
      jsx: true,
    },
    

Suggested fix
-------------

Since the Turbopack loader handles multiple file extensions, I think the parser configuration should ideally be selected from `modulePath`:

    parser: {
      syntax: /\.(ts|tsx|cts|mts)$/.test(modulePath)
        ? "typescript"
        : "ecmascript",
      jsx: /\.(jsx|tsx)$/.test(modulePath),
    },
    

This would result in:

| Extension | SWC parser |
| --- | --- |
| `.js` | ecmascript |
| `.jsx` | ecmascript + JSX |
| `.ts` | typescript |
| `.tsx` | typescript + JSX |
| `.cjs` / `.mjs` | ecmascript |
| `.cts` / `.mts` | typescript |

I have tested the explicit `jsx: true` workaround locally, and it makes both `next dev` and `next build` work correctly with Turbopack.

I hope this helps narrow down the issue. I'd be happy to provide any additional information or a minimal reproduction if needed.
Thanks again for maintaining the project and for taking a look!

Mad-Kat:
Closes: #638

The Turbopack loader ran SWC without a parser config, so `.js`, `.jsx`, `.mjs` and `.cjs` files containing JSX failed to parse. The Vite plugin had the opposite problem and always parsed as TSX regardless of the extension.

This adds a small shared helper that derives the SWC parser options from the file extension, mirroring how `next.js` handles it, and wires it into both loaders. Import attributes (`with { type: "json" }`) are now preserved through the transform as well.

### DIFF
--- a/packages/next-yak/loaders/lib/swcParserOptions.ts
+++ b/packages/next-yak/loaders/lib/swcParserOptions.ts
@@ -6,8 +6,8 @@ const typeScriptWithJsx = /\.tsx$/;
 /**
  * Derive the SWC parser configuration for a single file.
  *
- * JSX is enabled everywhere except in `.ts`, `.mts` and `.cts`,
- * where angle brackets are type assertions and generics rather than elements.
+ * JSX is not allowed in `.ts`, `.mts` and `.cts`
+ * therefore we must parse these as pure TS/JS and must not return JSX
  *
  * This is inspired by next.js own handling.
  */
--- a/packages/next-yak/loaders/lib/swcParserOptions.ts
+++ b/packages/next-yak/loaders/lib/swcParserOptions.ts
@@ -1,0 +1,24 @@
+import type { ParserConfig } from "@swc/core";
+
+const typeScriptWithoutJsx = /\.[mc]?ts$/;
+const typeScriptWithJsx = /\.tsx$/;
+
+/**
+ * Derive the SWC parser configuration for a single file.
+ *
+ * JSX is enabled everywhere except in `.ts`, `.mts` and `.cts`,
+ * where angle brackets are type assertions and generics rather than elements.
+ *
+ * This is inspired by next.js own handling.
+ */
+export function getSwcParserOptions(filename: string): ParserConfig {
+  // module ids may carry a query string (`/a/b.ts?v=1`)
+  const [path] = filename.split("?");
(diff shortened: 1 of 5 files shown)


packages/next-yak/loaders/lib/swcParserOptions.ts
```diff
@@ -0,0 +1,24 @@
+import type { ParserConfig } from "@swc/core";
+
+const typeScriptWithoutJsx = /\.[mc]?ts$/;
+const typeScriptWithJsx = /\.tsx$/;
+
+/**
+ * Derive the SWC parser configuration for a single file.
+ *
+ * JSX is enabled everywhere except in `.ts`, `.mts` and `.cts`,
+ * where angle brackets are type assertions and generics rather than elements.
```
jantimon:
```suggestion
 * JSX is not allowed in `.ts`, `.mts` and `.cts`
 * therefore we must parse these as pure TS/JS and must not return JSX
```

jantimon (approved):

PR #641, opened 2026-09-04
Update dependencies

Mad-Kat:
This updates all dependencies as far as possible with some exceptions:

- `@swc/core` needs to be in sync with `swc_core` and the bundled version in next.js
- `postcss-nested` (dev-dependency) can't be updated as next.js invocation of it can't handle the new ESM-only package
- `typescript` can't go to 7, as it doesn't yet have a programmatic API yet

Additionally I've changed all direct production dependencies to be a range instead of a hard pin, so that users can deduplicate transitive packages better


.github/workflows/autofix.yml
```diff
@@ -65,6 +65,12 @@ jobs:
       - name: Lint (oxlint)
         run: pnpm lint:fix
 
+      - name: Build next-yak and @yak/solid (needed by knip)
+        run: pnpm build
+
+      - name: Lint (knip)
```
jantimon, line 71: ❤️

.github/workflows/release.yml
```diff
@@ -71,21 +71,9 @@ jobs:
       - name: Verify npm version
         run: npm --version
 
-      # changesets/action runs `git push` directly, so it needs an authenticated remote.
-      # Done explicitly here (rather than via checkout's persist-credentials) so the
-      # credential is scoped to this step's needs and zizmor can audit the setup.
-      - name: Configure git auth for changesets push
-        env:
-          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
```
jantimon, line 79: ❤️

jantimon (approved):

PR #645, opened 2026-09-06
Server-render Solid in e2e, fail tests on console output, fix css prop hydration

jantimon:
Two things for the e2e suite and one bug they found. The runtime performance work for `@yak/solid` is stacked on top in #644.

#### Console output fails a test

Improved: now every test will fail on browser console errors and warnings and on page errors. `withTestEnv` collects them and throws after the test body, so an assertion failure is reported as itself with the console output printed next to it. A React hydration mismatch logs `console.error`, a Solid key miss logs a warning and leaves the element inert, so both show up on every bundler that server-renders. A case that provokes errors on purpose calls `testEnv.expectConsoleErrors("why")`; only the syntax-error HMR case does. `YAK_E2E_CONSOLE_SOFT=1` logs instead of failing, for collecting what a new bundler prints. A discovery run over all nine bundlers found no other console output.

#### `vite-solid` server-renders and hydrates

The Solid bundler rendered client-only with `render()`, so nothing checked a Solid server render. It now server-renders every case and hydrates it. `server.mjs` runs Vite in middleware mode for `dev`, HMR included, and serves the built client and server bundles for `start`. Each case gets its own page, server entry and client entry from the `[case-name]` templates, so a page carries only its own CSS. All Solid cases pass in dev and build, the HMR cases included.

Three new cases, each for React and Solid: `svg-namespace` for the four tag names HTML and SVG share (`a`, `script`, `style`, `title`) inside and outside an svg, plus a class update on the same element after a click; `theme` for the provider reaching style and attrs functions without leaking a `theme` attribute; `wrapped-component` for a styled component target receiving the class, the author's props and a `component` prop but no `$` props. With those the styled unit tests of `@yak/solid` are removed; the e2e suite is that runtime's test layer.

Solid moves to 2.0.0-rc.6 and `vite-plugin-solid` to its new name `@solidjs/vite-plugin`.

#### Fix: a css prop with a spread did not hydrate on Solid

Found by the suite on the first run. For `<button css={...} {...props} />` the transform emitted `__yak_mergeCssProp({ ...props }, css)`. The Solid compiler passes that call straight to `ssrElement` on the server, before the hydration key is taken, but wraps it in a callback on the client, after the element is claimed. The object spread read the `children` getter, which allocates a memo id, so the button's key drifted by one, the client created a detached element and the click handler never attached.

The compiler now emits the css value first and one argument per merged source in JSX order:

```tsx
<button class="a" {...props} style={s} css={css`color: green;`} />
// before
<button {...__yak_mergeCssProp({ class: "a", ...props, style: s }, css("yak1"))} />
// after
<button {...__yak_mergeCssProp(css("yak1"), { class: "a" }, props, { style: s })} />
```

The React helper merges the sources up front as before, so React behaviour is unchanged apart from the argument order. The Solid helper copies the sources' property descriptors without reading them and resolves `class` and `style` where they are read: on the client as getters the spread tracks, so a signal read by a source or by the css prop updates the attribute in place; on the server once, after `ssrElement` took the key. Twelve fixture directories change shape. The css-prop e2e case asserts the click, a spread's class joining the css prop's class, a later source replacing an earlier class, and a live interpolation, on every bundler.


docs/content/docs/how-does-it-work.mdx
```diff
@@ -243,7 +243,15 @@ const Elem = ({ on }) => <div className={on ? "input_Elem_m7uBBu" : "input_Elem_
...
-Only if the element has an existing `className` attribute or spread attributes the class names have to be merged at render time. In that case the plugin wraps the attributes with the small `__yak_mergeCssProp` runtime helper.
+Only if the element has an existing `className` attribute or spread attributes the class names have to be merged at render time. In that case the plugin hands the attributes to the small `__yak_mergeCssProp` runtime helper: the css value first, then one argument per merged source in JSX order.
+
+```tsx title="output"
+const Elem = (props) => (
+  <div {...__yak_mergeCssProp(css("input_Elem_m7uBBu"), { className: "card" }, props)} />
+);
+```
+
+Handing the sources over separately keeps the call itself free of reads. The Solid helper copies the property descriptors instead of spreading them, so no prop getter runs before the element takes its hydration key, and `class` and `style` stay live bindings that update in place.
```
jantimon:
```suggestion
```

.changeset/css-prop-sources.md
```diff
@@ -0,0 +1,7 @@
+---
+"next-yak": patch
+"yak-swc": patch
+"@yak/solid": patch
+---
+
+A css prop combined with a spread hydrates on Solid. The compiler now calls `__yak_mergeCssProp` with the css value first and one argument per merged source in JSX order instead of spreading them into one object at the call site. The Solid helper copies the sources' property descriptors without reading them and resolves `class` and `style` where they are read, so no prop getter runs before the element takes its hydration key, and the class binding stays live.
```
Mad-Kat:
```suggestion
Fix an issue where a css prop combined with a spread didn't hydrate on SolidJS.
```

docs/content/docs/folding.mdx
```diff
@@ -340,7 +340,7 @@ A conditional css prop keeps its condition; the branch toggles class names:
 </div>
 </SideBySide>
 
-The css prop keeps the render-time merge (`__yak_mergeCssProp`) when the element already has a `className`, spreads props, or carries a dynamic value (a css variable). A css prop that points at a `css` mixin declared elsewhere is a compile error, because the mixin carries no class name of its own.
+The css prop keeps the render-time merge (`__yak_mergeCssProp`) when the element already has a `className`, spreads props, or carries a dynamic value (a css variable). The helper takes the css value first and then one argument per merged source in JSX order, a run of `className` and `style` attributes as one object and a spread as the expression it was written as, so a spread is never copied into an object at the call site. A css prop that points at a `css` mixin declared elsewhere is a compile error, because the mixin carries no class name of its own.
```
Mad-Kat: I don't think this adds any value for this type of content

e2e/bundlers/vite-solid/package.json
```diff
@@ -3,18 +3,18 @@
   "private": true,
   "type": "module",
   "scripts": {
-    "dev": "cd .tmp && vite --port 5373",
-    "build": "cd .tmp && vite build",
-    "start": "cd .tmp && vite preview --port 5373"
+    "dev": "cd .tmp && node server.mjs",
```
Mad-Kat: I would prefer a server.ts script

packages/yak-solid/runtime/internals/mergeCssProp.ts
```diff
@@ -1,54 +1,87 @@
...
- * `}
- * {...{ style: { padding: "30px" }}}
- * />
+ * <button {...__yak_mergeCssProp(css("yak1"), { class: "a" }, props, { style: s })} />
+ * ```
+ * Optimized to not read the sources and therefore not invoke their Solid
+ * getters: the descriptors are copied untouched, and `class` and `style` are
+ * resolved by whoever reads them. A getter invoked here would run before the
+ * element takes its hydration key on the server but after it on the client,
+ * and a `children` getter that renders would then take its id on one side only.
```
Mad-Kat: I think this adds no value to read

Mad-Kat (approved):

docs/content/docs/folding.mdx
jantimon: Yes absolutely - this does not help to understand the idea and goes way to deep into detail

jantimon: Done: the sentence is gone, the paragraph reads as before.

e2e/bundlers/vite-solid/package.json
jantimon: Done: `server.mjs` is now `server.ts` (annotations only, Node runs it directly), and the `dev`/`start` scripts point at it.

packages/yak-solid/runtime/internals/mergeCssProp.ts
jantimon: Done: the paragraph is removed, the comment now only shows the compiled call.

PR #647, opened 2026-09-06
Check the e2e case layout

jantimon:
Stacked on #645. One script, `scripts/e2e-exists.mjs`, pins the layout every e2e case follows:

```
e2e/cases/<case>/index.tsx          the React case, required
e2e/cases/<case>/index.test.ts      the shared test, required
e2e/cases/<case>/<name>.solid.<ext> a Solid variant of <name>.<ext>, which must exist
```

A Solid variant replaces its React twin during assembly, so a variant without a twin would run on Solid only and never on React. Any other framework marker in a file name (`index.react.tsx`) is a typo the runner would silently ship to every bundler; the script rejects it. It uses Node's own `fs.globSync`, runs before `runDevTests.ts` and `runBuildTests.ts` start and as part of `pnpm lint`, and prints one line per problem. Changing the layout on purpose means changing this script, so it does not drift by accident.

The runtime PR #644 is rebased on top of this and gains the React twins its `reactive-props` case was missing.

### DIFF
--- a/e2e/README.md
+++ b/e2e/README.md
@@ -106,8 +106,9 @@ file. Variants therefore import siblings by their unmarked names, and the
 shared `index.test.ts` needs no changes. Tests can branch on
 `testEnv.framework` when the frameworks behave differently.
 
-A case without an `index.tsx` is exclusive to the framework of its variants:
-React bundlers skip it. Helper files in such a case need no `.solid.` marker.
+Every case has an `index.tsx`, an `index.test.ts`, and a React twin for each
+`.solid.` variant; `checkCaseLayout.ts` checks that layout when the runner
+starts, so a case cannot run on one framework only by accident.
 
 ## Adding a test case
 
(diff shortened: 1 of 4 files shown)


e2e/package.json
```diff
@@ -6,8 +6,8 @@
     ".": "./test-env.ts"
   },
   "scripts": {
-    "test": "node runDevTests.ts",
-    "test:build": "node runBuildTests.ts"
+    "test": "node ../scripts/e2e-exists.mjs && node runDevTests.ts",
```
Mad-Kat: do we want a `.ts` file instead?

Mad-Kat (approved):

e2e/package.json
jantimon: Done: the check is now `e2e/checkCaseLayout.ts`, called first thing by `runDevTests.ts` and `runBuildTests.ts`, and the three `package.json` scripts are back to what they were.

